PR #8, opened 2016-01-14
Bumping Json.Net version to 8.0.1

pranavkm:
cc @rynowak \ @davidfowl


pranavkm:
Mvc uses a higher version - I'm not super sure if we're vested in targeting the lowest version here.

rynowak:
Yep. 8.0.1 is our minimum for MVC. (Soon to be 8.0.2 when that's available).

PR #20, opened 2016-04-13
Migrate tests/samples/tools to portable

pakrym:
@davidfowl @anurse @pranavkm


anurse:
:shipit: on green builds

PR #21, opened 2016-04-21
Feature: Generate JsonPatchDocument from 2 objects (diff)

alanprot:
Do you guys think that it could be useful?
I' m using it here for audit for example.


alanprot:
PS: Code based in: https://github.com/khalidsalomao/SimpleHelpers.Net/blob/master/SimpleHelpers/ObjectDiffPatch.cs

Eilon:
Hi @alanprot , thanks for the PR, however at this time we are not looking to expand the functionality of the JsonPatch feature to include diff support.

PR #30, opened 2016-06-30
Updating to RTM builds of xunit and Moq

pranavkm:


javiercn:
:shipit:

test/Microsoft.AspNetCore.JsonPatch.Test/project.json
```diff
@@ -16,30 +17,15 @@
...
         },
-        "moq.netcore": "4.4.0-beta8",
-        "System.Diagnostics.Process": "4.1.0-*",
         "System.Diagnostics.TraceSource": "4.0.0-*"
       },
-      "imports": [
-        "dnxcore50",
-        "portable-net451+win8"
-      ]
+      "imports": "dotnet5.4"
```
javiercn: why do we need this import? Moq?

pranavkm: Yeah, the castle core package they are referencing is busted. But it gets us on RTM packages and off the custom builds we have so I guess it's worth the effort for now.

Issue #29, opened 2016-06-17
Honor JsonProperty when serializing a JsonPatchDocument.


KevinDockx:
An issue was logged on the issue list of Marvin.JsonPatch (on which AspNetCore.JsonPatch is based, which is why I'm logging this issue) regarding honoring JSON.NET's JsonProperty (https://github.com/KevinDockx/JsonPatch/issues/44#issuecomment-225681404).  

The AspNetCore version supports this when applying a patchDoc to a class that has properties decorated with [JsonProperty] attributes. Marvin.JsonPatch doesn't (PCL stuff), so that's not an issue for AspNetCore's version.  But it's the last comment of Amnesia1187 that's of importance, as it affects AspNetCore.JsonPatch. 

It's regarding creating a JsonPatchDocument on the consumer-side of things. Such a document is serialized to a Json list of operations, and the "path"-portion contains the property name the patch operation affects.  That part apparently does not respect JsonProperty values. 

So, if you create a JsonPatchDocument<MyClass>, and MyClass contains a property "Name" decorated with JsonProperty("RealName"), it will be serialized to path: "/name" instead of "realName". 

I feel it should serialize to realName.  Thoughts on this?

KevinDockx:
FYI, I fixed this issue in Marvin.JsonPatch.  I'd be happy to provide a PR with the required changes :)

Eilon:
Looks like https://github.com/aspnet/JsonPatch/pull/35 was merged, so closing this bug. Thanks!

PR #35, opened 2016-09-15
Fix for issue #29 (Honor JsonProperty when serializing a JsonPatchDocument)

KevinDockx:
Contains fix for issue #29 + unit tests.


test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
```diff
@@ -0,0 +1,113 @@
...
+using Xunit;
+
+namespace Microsoft.AspNetCore.JsonPatch.Test
+{
+    public class JsonPropertyTests
+    {
+        [Fact]
+        public void HonourJsonPropertyOnSerialization()
+        {
+            JsonPatchDocument<JsonPropertyDTO> patchDoc = new JsonPatchDocument<JsonPropertyDTO>();
```
rynowak: var

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
```diff
@@ -0,0 +1,113 @@
...
+            var deserialized =
+                JsonConvert.DeserializeObject<JsonPatchDocument<JsonPropertyWithAnotherNameDTO>>(serialized);
+
+            Assert.Equal(deserialized.Operations.First().path, "/anothername");
+        }
+
+        [Fact]
+        public void CanApplyToDifferentlyTypedClassWithPropertyMatchingJsonPropertyName()
+        {
+            JsonPatchDocument<JsonPropertyDTO> patchDocToSerialize =
```
rynowak: var

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            deserialized.ApplyTo(doc);
+
+            Assert.Equal(doc.AnotherName, "John");
+        }
+
+        [Fact]
+        public void CanApplyToSameTypedClassWithMatchingJsonPropertyName()
+        {
+            JsonPatchDocument<JsonPropertyDTO> patchDocToSerialize =
```
rynowak: var

src/Microsoft.AspNetCore.JsonPatch/Helpers/ExpressionHelpers.cs
```diff
@@ -51,10 +54,35 @@ private static string GetPath(Expression expr, bool firstTime)
...
+                        var jsonPropertyAttribute =
+                            memberExpression.Member.GetCustomAttributes(
+                            typeof(JsonPropertyAttribute), false);
+
+                        if (jsonPropertyAttribute.Count() > 0)
+                        {
+                            // get value
+                            var castedAttribrute = jsonPropertyAttribute.First() as JsonPropertyAttribute;
+                            return left + "/" + castedAttribrute.PropertyName;
+                        }
```
rynowak: Can you refactor the code for getting the property into a private method, it's basically a dupe between these two  cases

src/Microsoft.AspNetCore.JsonPatch/Helpers/ExpressionHelpers.cs
```diff
@@ -51,10 +54,35 @@ private static string GetPath(Expression expr, bool firstTime)
...
                     {
                         var left = GetPath(memberExpression.Expression, false);
+
+                        // if there's a JsonProperty attribute, we must return the PropertyName
+                        // from the attribute rather than the member name 
+                        var jsonPropertyAttribute =
+                            memberExpression.Member.GetCustomAttributes(
+                            typeof(JsonPropertyAttribute), false);
+
+                        if (jsonPropertyAttribute.Count() > 0)
```
rynowak: `GetCustomAttributes` returns an `object[]`, it's not necessary to use linq here.

src/Microsoft.AspNetCore.JsonPatch/Helpers/ExpressionHelpers.cs
```diff
@@ -51,10 +54,35 @@ private static string GetPath(Expression expr, bool firstTime)
                     if (ContinueWithSubPath(memberExpression.Expression.NodeType, false))
                     {
                         var left = GetPath(memberExpression.Expression, false);
+
+                        // if there's a JsonProperty attribute, we must return the PropertyName
+                        // from the attribute rather than the member name 
+                        var jsonPropertyAttribute =
+                            memberExpression.Member.GetCustomAttributes(
+                            typeof(JsonPropertyAttribute), false);
```
rynowak: this should be `inherited: true`. `JsonPropertyAttribute` supports inheritance

src/Microsoft.AspNetCore.JsonPatch/Helpers/ExpressionHelpers.cs
```diff
@@ -51,10 +54,35 @@ private static string GetPath(Expression expr, bool firstTime)
...
+
+                        if (jsonPropertyAttribute.Count() > 0)
+                        {
+                            // get value
+                            var castedAttribrute = jsonPropertyAttribute.First() as JsonPropertyAttribute;
```
rynowak: This should just be `(JsonPropertyAttribute)jsonPropertyAttribute[0]` - it should be a c-style cast because it will never fail

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyDTO.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.AspNetCore.JsonPatch.Test
```
rynowak: You can drop `.Test` from all of these namespaces

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
```diff
@@ -0,0 +1,113 @@
...
+using Newtonsoft.Json;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.AspNetCore.JsonPatch.Test
+{
+    public class JsonPropertyTests
+    {
+        [Fact]
+        public void HonourJsonPropertyOnSerialization()
```
rynowak: We tend to name tests like `WhenIDoSomething_AnotherThingHappens_AndAnotherThing` - this would be like `Add_WithExpression_RespectsJsonPropertyName_ForModelProperty`

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
```diff
@@ -0,0 +1,113 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.AspNetCore.JsonPatch.Test
+{
+    public class JsonPropertyTests
```
rynowak: These are really tests for `JsonPatchDocument` so this should be `JsonPatchDocumentTest` or if you prefer `JsonPatchDocumentJsonPropertyAttributeTest`

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
```diff
@@ -0,0 +1,113 @@
...
+            JsonPatchDocument<JsonPropertyDTO> patchDoc = new JsonPatchDocument<JsonPropertyDTO>();
+            patchDoc.Add(p => p.Name, "John");
+
+            var serialized = JsonConvert.SerializeObject(patchDoc);
+            // serialized value should have "AnotherName" as path
+            // deserialize to a JsonPatchDocument<JsonPropertyWithAnotherNameDTO> to check
+            var deserialized =
+                JsonConvert.DeserializeObject<JsonPatchDocument<JsonPropertyWithAnotherNameDTO>>(serialized);
+
+            Assert.Equal(deserialized.Operations.First().path, "/anothername");
```
rynowak: See https://github.com/aspnet/Home/wiki/Engineering-guidelines#unit-test-structure

KevinDockx:
That's indeed a valid point (reflection), but I'm not sure if it's possible to use JsonContract in this case (if it is, I'm guessing it'll require quite a few more changes).  Anyway, for now I'll fix the unit test comments & start thinking about a different approach :)

rynowak:
> Anyway, for now I'll fix the unit test comments & start thinking about a different approach :)

Yeah I think we want to get this fix in and consider a follow up.

src/Microsoft.AspNetCore.JsonPatch/Helpers/ExpressionHelpers.cs
KevinDockx: Can't use indexing on IEnum<T> apparently, but I changed the cast to a c-type cast.

src/Microsoft.AspNetCore.JsonPatch/Helpers/ExpressionHelpers.cs
KevinDockx: Done, tests added.

src/Microsoft.AspNetCore.JsonPatch/Helpers/ExpressionHelpers.cs
KevinDockx: I should've seen that :) done

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyDTO.cs
KevinDockx: Done

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
KevinDockx: Done

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
KevinDockx: Ok, fixed test naming

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
KevinDockx: done

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
KevinDockx: Done

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
KevinDockx: Done

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyTests.cs
KevinDockx: Done

src/Microsoft.AspNetCore.JsonPatch/Helpers/ExpressionHelpers.cs
```diff
@@ -93,6 +70,26 @@ private static string GetPath(Expression expr, bool firstTime)
...
+        private static string GetPropertyNameFromMemberExpression(MemberExpression memberExpression)
+        {
+            // if there's a JsonProperty attribute, we must return the PropertyName
+            // from the attribute rather than the member name 
+            var jsonPropertyAttribute =
+                memberExpression.Member.GetCustomAttributes(
+                typeof(JsonPropertyAttribute), true);
+
+            // to object[] to avoid LINQ
+            var jsonPropertyAttributeAsObjectArray = (object[])jsonPropertyAttribute;
```
rynowak: This is not OK :laughing: - we shouldn't do a downcast here because this is a framework implementation detail.

You should be able to use https://msdn.microsoft.com/en-us/library/hh138307(v=vs.110).aspx here and avoid linq and downcasting

KevinDockx: Aha, now I get it ;-) Couldn't avoid LINQ without the downcast, but I missed the method to get one attribute instead of having to go through getting the list & then the first one from the list.  Should be fixed now :)

rynowak:
:shipit: - looks good

rynowak:
Can you squash? I'd like to try out this fancy new rebase + merge butter

rynowak:
NVM I took care of it. Thanks!

PR #36, opened 2016-09-24
[Fixes #33] Dictionary operations fail due to contract issues

### LINKED ISSUE #33: Patching fails for non 'object' value type dictionaries
Given a model like this:

```csharp
public class DemoModel
{
  public DemoModel()
  {
    this.Numbers = new Dictionary<string, int>();
  }

  public IDictionary<string, int> Numbers { get; private set; }
}
```

When JSON serialized, the object looks like this:

```json
{
  "Numbers": {
    "one":1,
    "two":2
  }
}
```

If I try to create a patch document and apply it, it fails:

```csharp
[Fact]
public void DictionaryTest()
{
  var model = new DemoModel();
  model.Numbers["one"] = 1;
  model.Numbers["two"] = 2;

  var patch = new JsonPatchDocument();
  patch.Operations.Add(new Operation("add", "/Numbers/three", null, 3));
  patch.ApplyTo(model);

  Assert.Equal(3, model.Numbers.Count);
  Assert.Equal(1, model.Numbers["one"]);
  Assert.Equal(2, model.Numbers["two"]);
  Assert.Equal(3, model.Numbers["three"]);
}
```

The test fails on the `patch.ApplyTo(model)` line:

```
at Microsoft.AspNetCore.JsonPatch.Helpers.ObjectTreeAnalysisResult..ctor(Object objectToSearch, String propertyPath, IContractResolver contractResolver)
   at Microsoft.AspNetCore.JsonPatch.Adapters.ObjectAdapter.Add(String path, Object value, Object objectToApplyTo, Operation operationToReport)
   at Microsoft.AspNetCore.JsonPatch.Adapters.ObjectAdapter.Add(Operation operation, Object objectToApplyTo)
   at Microsoft.AspNetCore.JsonPatch.JsonPatchDocument.ApplyTo(Object objectToApplyTo, IObjectAdapter adapter)
   at DemoProject.DictionaryTest() in C:\path\to\Fixture.cs:line 73
Result Message:	Unable to cast object of type 'Newtonsoft.Json.Serialization.JsonDictionaryContract' to type 'Newtonsoft.Json.Serialization.JsonObjectContract'.
```

Is this a known issue? Or is there a workaround?

kichalla:
@dougbu @rynowak


src/Microsoft.AspNetCore.JsonPatch/Helpers/ConversionHelper.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.AspNetCore.JsonPatch.Helpers
```
kichalla: There are existing classes which are internal and use this namespace. I left it as it is currently so that we can clean up all these in a separate task.

src/Microsoft.AspNetCore.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
```diff
@@ -127,12 +138,15 @@ internal class ObjectTreeAnalysisResult
             // case, it's valid for add if there's 1 item left in the propertyPathTree.
             //
             // it can also be a property info.  In that case, if there's nothing left in the path
-            // tree we're at the end, if there's one left we can try and set that. 
-            if (targetObject is IDictionary<string, object>)
+            // tree we're at the end, if there's one left we can try and set that.
+            var idictionaryType = ClosedGenericMatcher.ExtractGenericInterface(
```
kichalla: There is duplication of logic in this method which we can cleanup later (I have a separate set of changes which are bigger and probably need some discussion before we agree to proceed with them), but for now I tried to do minimal set of changes to fix this issue.

src/Microsoft.AspNetCore.JsonPatch/Helpers/DictionaryWrapper.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        public bool ContainsCaseInsensitiveKey(object key)
+        {
+            var foundKey = GetKeyUsingCaseInsensitiveSearch(key);
+            return _targetDictionary.ContainsKey(foundKey);
+        }
+
+        private TKey GetKeyUsingCaseInsensitiveSearch(object key)
+        {
+            // Example: a Guid key
```
kichalla: Example json patch requests for reference:

```
[
     { "op": "add", "path": "/baz", "value": "qux" },
     { "op": "add", "path": "/foo/1", "value": "qux" }
   ]
```
Spec: https://tools.ietf.org/html/rfc6902

test/Microsoft.AspNetCore.JsonPatch.Test/ObjectAdapterTests.cs
```diff
@@ -1733,5 +1733,296 @@ public void MoveToEndOfListWithSerialization()
             Assert.Equal(0, doc.IntegerValue);
             Assert.Equal(new List<int>() { 1, 2, 3, 5 }, doc.IntegerList);
         }
+
+        private class Class6
```
dougbu: Add tests of case-insensitive dictionary keys to this class.

kichalla: There are already existing test cases which cover this

src/Microsoft.AspNetCore.JsonPatch/project.json
```diff
@@ -21,23 +21,14 @@
...
-        "System.Globalization": "4.0.11-*",
-        "System.Linq": "4.1.0-*",
-        "System.Reflection.Extensions": "4.0.1-*",
-        "System.Resources.ResourceManager": "4.0.1-*",
-        "System.Runtime.Extensions": "4.1.0-*",
-        "System.Runtime.Serialization.Primitives": "4.1.1-*",
-        "System.Text.Encoding.Extensions": "4.0.11-*"
-      }
-    }
+    "netstandard1.3": { }
```
kichalla: This is because of the following line: https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.ClosedGenericMatcher.Sources/ClosedGenericMatcher.cs#L68

And the package that has it is in 1.3 http://packagesearch.azurewebsites.net/?q=GetInterfaces

src/Microsoft.AspNetCore.JsonPatch/Helpers/DictionaryWrapper.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        public void SetValueForCaseInsensitiveKey(object key, object value)
+        {
+            var foundKey = GetKeyUsingCaseInsensitiveSearch(key);
+
+            // If the target object is an ExpandoObject, then the type of the value in the key-value pair is 'object'
+            // and so we cannot deserialize the raw json to a particular type.
+            // But if a value for a key already exists, then we try to figure out the type and deserialize the value to
+            // this type.
+            if (useDynamic && _targetDictionary.ContainsKey(foundKey) && _targetDictionary[foundKey] != null)
```
kichalla: All of your points here are valid. However, looks like I actually do not need this code as its been handled already elsewhere (This is a remnant from a bigger change that I have been working on).

> Am I correct this code will use a JObject value if the key doesn't exist in the dictionary? If so, should that case instead be an error?

This case is not handled in existing code. Let's take it offline.

src/Microsoft.AspNetCore.JsonPatch/Helpers/DictionaryWrapper.cs
```diff
@@ -0,0 +1,100 @@
...
+        private readonly bool useDynamic;
+
+        public DictionaryWrapper(IDictionary<TKey, TValue> targetDictionary)
+        {
+            _targetDictionary = targetDictionary;
+            KeyType = typeof(TKey);
+            ValueType = typeof(TValue);
+
+            // For the case when target dictionary is an ExpandoObject (which implements IDictionary<string, object>)
+            useDynamic = (KeyType == typeof(string)) && (ValueType == typeof(object));
```
kichalla: As mentioned in my other comment, this code is not longer required as its being handled elsewhere, but your comment is valid though.

src/Microsoft.AspNetCore.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
kichalla: I cannot use this as I have a similar name in this method.

src/Microsoft.AspNetCore.JsonPatch/Helpers/DictionaryWrapper.cs
```diff
@@ -0,0 +1,100 @@
...
+                if (!conversionResult.CanBeConverted)
+                {
+                    throw new JsonPatchException(
+                        message: Resources.FormatInvalidDictionaryKeyFormat(key, KeyType.FullName),
+                        innerException: null);
+                }
+                return CastTo<TKey>(conversionResult.ConvertedInstance);
+            }
+
+            var keyToFind = (string)key;
```
kichalla: > Can't key be a non-string but have a type conversion to string?

Not sure what you mean here, but why would a key be converted to a string by a dictionary? isn't hashcode used to store it.

kichalla:
🆙 📅

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -1005,20 +1005,6 @@ private void LogError(JsonPatchError jsonPatchError)
             }
         }
 
-        private ConversionResult ConvertToActualType(Type propertyType, object value)
```
rynowak: There's gotta be a better way to do this.

src/Microsoft.AspNetCore.JsonPatch/Helpers/CaseInsensitiveDictionary.cs
```diff
@@ -0,0 +1,79 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.JsonPatch.Exceptions;
+
+namespace Microsoft.AspNetCore.JsonPatch.Helpers
+{
+    internal class CaseInsensitiveDictionary<TKey, TValue> : ICaseInsensitiveDictionary
```
rynowak: How is this different than `new Dictionary<>(StringComparer.OrdinalIgnoreCase)`?

src/Microsoft.AspNetCore.JsonPatch/Helpers/ConversionResultProvider.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.AspNetCore.JsonPatch.Helpers
+{
+    internal static class ConversionResultProvider
```
rynowak: `Helpers` is the first clue that something is wrong here. 

1. We should find a better way to do this conversion - for instance this should use `JsonContract`/`JsonConverter` and not do serialization.
2. The usability of this is wrong, this should be a `bool TryConvert(Type, object, out object)`

src/Microsoft.AspNetCore.JsonPatch/Helpers/ICaseInsensitiveDictionary.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.JsonPatch.Helpers
+{
+    internal interface ICaseInsensitiveDictionary
```
rynowak: This smells fishy - let's talk through why this is needed and what better solutions we might find

src/Microsoft.AspNetCore.JsonPatch/project.json
```diff
@@ -21,22 +21,17 @@
...
+    "Microsoft.Extensions.ClosedGenericMatcher.Sources": {
+      "type": "build",
+      "version": "1.1.0-*"
+    },
     "Newtonsoft.Json": "9.0.1"
   },
   "frameworks": {
-    "netstandard1.1": {
+    "net451": {},
+    "netstandard1.3": {
```
dougbu: Changing the supported frameworks remains a breaking change.

@kichalla you mentioned this had something to do w/ `ClosedGenericMatcher`. Could the requirements of that package be reduced?

@Eilon thoughts?

test/Microsoft.AspNetCore.JsonPatch.Test/project.json
```diff
@@ -20,7 +20,7 @@
         "System.Diagnostics.TraceSource": "4.0.0-*"
       }
     },
-    "net451": {
+    "net46": {
```
dougbu: Why?

test/Microsoft.AspNetCore.JsonPatch.Test/ObjectAdapterTests.cs
dougbu: Please point me to an example or two. I'm not seeing such tests.

src/Microsoft.AspNetCore.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
dougbu: Come up with another name; one that doesn't start w/ `idictionary` or `iDictionary`.

src/Microsoft.AspNetCore.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
```diff
@@ -42,19 +44,28 @@ internal class ObjectTreeAnalysisResult
             // we've now got a split up property tree "base/property/otherproperty/..."
             int lastPosition = 0;
             object targetObject = objectToSearch;
+            ICaseInsensitiveDictionary dictionaryWrapper = null;
```
dougbu: Name no longer means anything. Suggest `dictionary`.

src/Microsoft.AspNetCore.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
```diff
@@ -42,19 +44,28 @@ internal class ObjectTreeAnalysisResult
...
+                    typeof(IDictionary<,>));
+                if (dictionaryType != null)
                 {
-                    // find the value in the dictionary                   
-                    if (dictionary.ContainsCaseInsensitiveKey(propertyPathTree[i]))
+                    var keyType = dictionaryType.GetTypeInfo().GenericTypeArguments[0];
+                    var valueType = dictionaryType.GetTypeInfo().GenericTypeArguments[1];
+                    var concreteType = typeof(CaseInsensitiveDictionary<,>).MakeGenericType(keyType, valueType);
+                    var wrapper = (ICaseInsensitiveDictionary)Activator.CreateInstance(concreteType, args: targetObject);
+                    dictionaryWrapper = wrapper;
```
dougbu: Why does `wrapper` exist as a separate variable?

test/Microsoft.AspNetCore.JsonPatch.Test/ObjectAdapterTests.cs
kichalla: https://github.com/aspnet/jsonpatch/blob/dev/test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/AddOperationTests.cs#L387
https://github.com/aspnet/jsonpatch/blob/dev/test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/AddOperationTests.cs#L382

test/Microsoft.AspNetCore.JsonPatch.Test/project.json
kichalla: oops...this wasn't supposed to be changed

src/Microsoft.AspNetCore.JsonPatch/project.json
kichalla: I thought about the version, if JsonPatch was used along with MVC then it should not be a problem as MVC requires netstandard1.6, and if its was used outside of MVC then if we consider HttpAbstractions, it requires netstandard1.3. Considering these this change should be fine?

src/Microsoft.AspNetCore.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
kichalla: I tried 😄 . Suggestions?

dougbu: Looks like `ReferenceEquals(targetObject, dictionaryWrapper)` (w/ new name for `dictionaryWrapper`) would work fine for the condition below. No need for another `ClosedGenericMatcher.ExtractGenericInterface()` call.

src/Microsoft.AspNetCore.JsonPatch/project.json
Eilon: Hmm interesting. I suppose in a sense, all that matters is the *effective* target framework support.

kichalla:
Closing this PR as its not relevant anymore. The new PR is https://github.com/aspnet/JsonPatch/pull/40

PR #40, opened 2016-09-28
[Fixes #33] Patching fails for non 'object' value type dictionaries

### LINKED ISSUE #33: Patching fails for non 'object' value type dictionaries

kichalla:
@rynowak @dougbu 

This is a fairly big change, but I can go through it with you and explain. All the existing tests pass and I have added some new coverage too.


kichalla:
@dougbu @rynowak The latest commit has changes related to our discussions yesterday. Let me know how it looks. (All the tests pass with these changes, but since this new commit creates new classes, I yet need to write unit tests for them, but would like know if this looks the way you expect)

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -139,226 +136,27 @@ public void Add(Operation operation, object objectToApplyTo)
...
-                else
-                {
-                    // New property - add it.  
-                    treeAnalysisResult.Container.Add(treeAnalysisResult.PropertyPathInParent, value);
-                }
+                var patchObject = ObjectVisitor.Visit(context);
+                patchObject.Add(value);
             }
-            else
+            catch (JsonPatchException exception)
```
rynowak: This could also use `catch (...) when (...)` which would be an improvement in clarity as well as provide essential PranavPoints.

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryObjectVisitor.cs
```diff
@@ -0,0 +1,41 @@
...
+            var _dictionary = (IDictionary)context.TargetObject;
+
+            // Since PathSegment is struct and the dictionary's key is of type 'object', convert it explicitly to
+            // string, otherwise look up will fail.
+            var key = pathSegment.ToString();
+
+            // Example: /USStatesProperty/WA
+            if (pathSegment.IsFinal)
+            {
+                return new PatchDictionaryObject(_dictionary, pathSegment, context.Operation);
```
rynowak: This is different than what we talked about

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryObjectVisitor.cs
```diff
@@ -0,0 +1,41 @@
...
+            }
+            else if (_dictionary.Contains(key))
+            {
+                // Example path: "/Customers/101/Address/Zipcode" and
+                // let's say the current path segment is "101"
+                var newTargetObject = _dictionary[key];
+                context.SetNewTargetObject(newTargetObject);
```
rynowak: Put the state on the Visitor, not on a Context.

src/Microsoft.AspNetCore.JsonPatch/Internal/ExpandoObjectDictionaryExtensions.cs
```diff
@@ -1,13 +1,13 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Generic;
 
-namespace Microsoft.AspNetCore.JsonPatch.Helpers
+namespace Microsoft.AspNetCore.JsonPatch.Internal
```
rynowak: :heart_eyes:

kichalla:
🆙 📅

src/Microsoft.AspNetCore.JsonPatch/Internal/ObjectVisitor.cs
```diff
@@ -0,0 +1,76 @@
...
+            if (target == null)
+            {
+                adapter = null;
+                errorMessage = null;
+                return false;
+            }
+
+            adapter = SelectAdapater(target);
+
+            // Traverse till the penultimate segment to get the target object and adapter
```
rynowak: 22 RyanPoints awarded

test/Microsoft.AspNetCore.JsonPatch.Test/ErrorMessageFormats.cs
```diff
@@ -0,0 +1,31 @@
+﻿
+
+namespace Microsoft.AspNetCore.JsonPatch
+{
+    public static class ErrorMessageFormats
```
rynowak: We don't generally do things like this, our general practice is to just put the text in the test.  The **reason** for this is so that you have to double check that the message makes sense in all the cases it is used when making changes. Doing things this way defeats that.

This is one of @Eilon 's rules

src/Microsoft.AspNetCore.JsonPatch/Internal/ObjectVisitor.cs
Eilon: -1 EilonPoint for using "till" instead of "til" (or " 'til ")

Issue #44, opened 2016-10-08
Missing support for JSON Pointer Escaping


sandorfr:
The following JSON is valid:

``` json
{
  "foo/bar~" : "a value"
}
```

But currently the library does not support `~` and `/` as stated under [JSON Pointer section](http://jsonpatch.com/), so this can't be patched.

I'll open a pull request with a fix proposal.

Eilon:
@kichalla do you have time to review the PR for this today? I'd like to get this into 1.1.0.

kichalla:
@Eilon Sorry I actually didn't see the notification about this until few minutes back. Taking a look now.

Eilon:
Looks like you merged this, so closing. Thanks!

PR #45, opened 2016-10-08
Escaping support for JSON pointers

sandorfr:
- Add Escaping support in ParsedPath class:
- Add Complex Name DTO requiring escaping
- Add Unit Test against that DTO

Addresses #44

### DIFF
--- a/src/Microsoft.AspNetCore.JsonPatch/Internal/ParsedPath.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/Internal/ParsedPath.cs
@@ -83,6 +83,11 @@ namespace Microsoft.AspNetCore.JsonPatch.Internal
                 }
             }
 
+            if (sb.Length > 0)
+            {
+                strings.Add(sb.ToString());
+            }
+
             return strings.ToArray();
         }
     }
--- a/src/Microsoft.AspNetCore.JsonPatch/Internal/ParsedPath.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/Internal/ParsedPath.cs
@@ -1,9 +1,11 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.JsonPatch.Exceptions;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using System.Text.RegularExpressions;
 
 namespace Microsoft.AspNetCore.JsonPatch.Internal
@@ -14,8 +16,6 @@ namespace Microsoft.AspNetCore.JsonPatch.Internal
 
         private readonly string[] _segments;
 
-        private static readonly Regex escapingRegex = new Regex("~(?<code>0|1)", RegexOptions.Compiled);
-
         public ParsedPath(string path)
         {
             if (path == null)
@@ -23,10 +23,7 @@ namespace Microsoft.AspNetCore.JsonPatch.Internal
                 throw new ArgumentNullException(nameof(path));
             }
 
-            _segments = path
-                .Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries)
-                .Select(segment => escapingRegex.Replace(segment, (match) => match.Groups["code"].Value == "0" ? "~" : "/"))
-                .ToArray();
+            _segments = ParsePath(path);
         }
 
         public string LastSegment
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.JsonPatch/Internal/ParsedPath.cs
```diff
@@ -12,14 +14,19 @@ public struct ParsedPath
...
         {
             if (path == null)
             {
                 throw new ArgumentNullException(nameof(path));
             }
 
-            _segments = path.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+            _segments = path
+                .Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries)
+                .Select(segment => escapingRegex.Replace(segment, (match) => match.Groups["code"].Value == "0" ? "~" : "/"))
```
Eilon: Can we change this to not use regex? It's a pretty big hammer (and super slow and inefficient), so it might be nicer to either use `string.Split`, or do a more manual conversion. I think the code is as simple as `<path>.Replace("~1", "/").Replace("~0", "~")`. Note that the order of these operations is significant because you don't want to double-decode the `~` character.

sandorfr: Hi, I've quickly wrote a [console app](https://github.com/sandorfr/random-stuff/blob/master/JsonPointer/ParseJsonPointerTest/src/ParseJsonPointerTest/Program.cs) to test two different approaches : 
Replace is definitely better than Regex. 

However, a custom implementation reducing memory allocation is slightly more performant than Split + replace. It also gives us a chance to catch bad escaping sequences.

```
RUN 1:
ParsePathWithSplitAndRegex :329 ms.
ParsePathWithSplitAndReplace :146 ms.
ParsePathWithCustomImplementation :133 ms.
RUN 2:
ParsePathWithSplitAndRegex :263 ms.
ParsePathWithSplitAndReplace :82 ms.
ParsePathWithCustomImplementation :84 ms.
RUN 3:
ParsePathWithSplitAndRegex :154 ms.
ParsePathWithSplitAndReplace :109 ms.
ParsePathWithCustomImplementation :119 ms.
RUN 4:
ParsePathWithSplitAndRegex :205 ms.
ParsePathWithSplitAndReplace :138 ms.
ParsePathWithCustomImplementation :97 ms.
RUN 5:
ParsePathWithSplitAndRegex :229 ms.
ParsePathWithSplitAndReplace :87 ms.
ParsePathWithCustomImplementation :69 ms.
```

Let met know what you think.

sandorfr: Fixed a call to a mysterious Enumerable.Append to List.Add and it's now obvious that the custom implementation is giving better results, So I'm pushing this approach to the branch.
```
RUN 1:
ParsePathWithSplitAndRegex :214 ms.
ParsePathWithSplitAndReplace :77 ms.
ParsePathWithCustomImplementation :27 ms.
RUN 2:
ParsePathWithSplitAndRegex :132 ms.
ParsePathWithSplitAndReplace :74 ms.
ParsePathWithCustomImplementation :31 ms.
RUN 3:
ParsePathWithSplitAndRegex :168 ms.
ParsePathWithSplitAndReplace :79 ms.
ParsePathWithCustomImplementation :26 ms.
RUN 4:
ParsePathWithSplitAndRegex :129 ms.
ParsePathWithSplitAndReplace :67 ms.
ParsePathWithCustomImplementation :28 ms.
RUN 5:
ParsePathWithSplitAndRegex :128 ms.
ParsePathWithSplitAndReplace :79 ms.
ParsePathWithCustomImplementation :27 ms.
```

src/Microsoft.AspNetCore.JsonPatch/Internal/ParsedPath.cs
```diff
@@ -1,8 +1,12 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.JsonPatch.Exceptions;
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
```
Eilon: Can remove this namespace now 😄

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyComplexNameDTO.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.AspNetCore.JsonPatch
+{
+    public class JsonPropertyComplexNameDTO
+    {
+        [JsonProperty("foo/bar~")]
```
Eilon: It would be good to add some more cases that require escaping, such as:

Valid cases:

1. `foo/bar~`
2. `foo/bar~0baz`
3. `foo/bar~1baz`
4. `foo/bar~0/~0/~1~1/~0~0/baz`
5. `~0~1foo`

And perhaps more error cases as well:

1. `~`
2. `foo~`
3. `foo~2bar`

src/Microsoft.AspNetCore.JsonPatch/Internal/ParsedPath.cs
sandorfr: System.Text is needed for the StringBuilder, I removed the other ones.

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPropertyComplexNameDTO.cs
sandorfr: I've added a two theories to test directly the ParsedPath class. I also added another property on this test.

I've also added a test involving dictionaries (which was my initial issue actually).

sandorfr: `foo/bar~` would be a valid property name but an `invalid escaped path`

Eilon:
@kichalla do you have time to review this today? I'd like to get this into 1.1.0.

kichalla:
Merged with commit: fe2453b93abe4f0bfbfade7abcb737fe10ed4040

Thanks for the PR!

sandorfr:
My pleasure :)

PR #46, opened 2016-10-31
Handle exceptions for invalid operation types

kichalla:
Issue in MVC: https://github.com/aspnet/Mvc/issues/5463


src/Microsoft.AspNetCore.JsonPatch/Internal/JsonPatchErrorHelper.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.JsonPatch.Exceptions;
+
+namespace Microsoft.AspNetCore.JsonPatch.Internal
+{
+    internal static class JsonPatchErrorHelper
```
rynowak: This is gross, it was better before

src/Microsoft.AspNetCore.JsonPatch/Operations/OperationBase.cs
```diff
@@ -13,7 +14,12 @@ public OperationType OperationType
         {
             get
             {
-                return (OperationType)Enum.Parse(typeof(OperationType), op, true);
+                OperationType result;
+                if (!Enum.TryParse<OperationType>(op, ignoreCase: true, result: out result))
+                {
+                    throw new JsonPatchException($"Invalid JsonPatch operation '{op}'.", innerException: null);
```
rynowak: Resources?

src/Microsoft.AspNetCore.JsonPatch/Operations/OperationBase.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNetCore.JsonPatch.Exceptions;
```
rynowak: Oooh is this something we already have 😢

src/Microsoft.AspNetCore.JsonPatch/Operations/OperationBase.cs
```diff
@@ -13,7 +14,12 @@ public OperationType OperationType
         {
             get
             {
-                return (OperationType)Enum.Parse(typeof(OperationType), op, true);
+                OperationType result;
+                if (!Enum.TryParse<OperationType>(op, ignoreCase: true, result: out result))
```
rynowak: is `ignoreCase` right for JSON patch?

kichalla: Good question. I couldn't find any information about case sensitiveness in JsonPatch spec. Since  Json.net deserialization in general is case INsensitive, I believe this is fine? also it could be a breaking change to change this behavior now as we were ignoring case earlier too.

src/Microsoft.AspNetCore.JsonPatch/Internal/JsonPatchErrorHelper.cs
kichalla: We need to report error in 3 classes, so moved it out a helper, so any suggestions or ok to keep it like this?

kichalla:
🆙 📅

src/Microsoft.AspNetCore.JsonPatch/Internal/JsonPatchErrorHelper.cs
rynowak: Using the word helper anywhere in a PR is a surefire way to get me to object to it. 

I'd suggest something like this.

```
public static class ErrorReporter
{
    public static readonly Action<JsonPatchError> Default = (error) => throw new JsonPatchException(error);
}
```

Then everywhere else is just 
```
errorReporter = errorReporter ?? ErrorReporter.Default
```

kichalla:
🆙 📅

kichalla:
81931e75d48370cf9163254523b1f6c4bcc92acb

PR #51, opened 2016-12-30
[Fixes #50] JsonPatchDocument.Replace() yields invalid path when [JsonProperty] is used (1.1.0)

### LINKED ISSUE #50: JsonPatchDocument.Replace() yields invalid path when [JsonProperty] is used (1.1.0)
### Title
In `1.1.0`, `JsonPatchDocument<T>.Replace()` creates `Operation<T>` with invalid `path`, when `[JsonProperty]` from `Newtonsoft.Json` is used.

### Functional impact
This is regression from `"Microsoft.AspNetCore.JsonPatch": "1.0.0"` and breaks existing behavior.

### Minimal repro steps

`global.json`:
```
  "sdk": {
    "version": "1.0.0-preview2-1-003177"
  }
```

`project.json`:
```
{
  "version": "1.0.0-*",
  "buildOptions": {
    "emitEntryPoint": true
  },

  "dependencies": {
    "Microsoft.AspNetCore.JsonPatch": "1.1.0",
    "Microsoft.NETCore.App": {
      "type": "platform",
      "version": "1.1.0"
    },
    "Newtonsoft.Json": "9.0.1"
  },

  "frameworks": {
    "netcoreapp1.0": {
      "imports": "dnxcore50"
    }
  }
}

```

`Program.cs`:
```
using System.Diagnostics;
using System.Linq;
using Microsoft.AspNetCore.JsonPatch;
using Newtonsoft.Json;

namespace JsonPatchReplaceIssue
{
    public class Model
    {
        [JsonProperty]
        public string[] ArrayOfStringsOrNull { get; set; }
    }

    public class Program
    {
        public static void Main(string[] args)
        {
            var patch = new JsonPatchDocument<Model>()
                .Replace(m => m.ArrayOfStringsOrNull, new[]
                {
                    "yo"
                });

            var operation = patch.Operations.Single();

            Debug.Assert(operation.path == $"/{nameof(Model.ArrayOfStringsOrNull).ToLowerInvariant()}",
                $"No it ain't! It is '{operation.path}'");
        }
    }
}
```

### Expected result
The `Operation<T>.path` should be derived from the property in the expression. Expected `path` to be `"/arrayofstringsornull"` in this case.

### Actual result
Actual `path` is `"/"`, which is incorrect.

### Further technical details
Regression from `"Microsoft.AspNetCore.JsonPatch": "1.0.0"` with `.NET Core 1.0`.
I have tested:
- Change the `JsonPatch` dependency in `project.json` to `1.0.0` and it works as expected.
- Remove `[JsonProperty]` and it works as expected.

### Question
Am I doing it wrong? I'm assuming that `Newtonsoft.Json`'s attributes are supported.

kichalla:
…nProperty] is used (1.1.0)

@rynowak


kichalla:
@pranavkm

src/Microsoft.AspNetCore.JsonPatch/Internal/ExpressionHelpers.cs
```diff
@@ -69,21 +69,12 @@ private static string GetPath(Expression expr, bool firstTime)
...
-                typeof(JsonPropertyAttribute), true);
-
-            if (jsonPropertyAttribute == null)
-            {
-                return memberExpression.Member.Name;
-            }
-            // get value
-            var castedAttribute = (JsonPropertyAttribute)jsonPropertyAttribute;
-            return castedAttribute.PropertyName;
+            var jsonObjectContract = (JsonObjectContract)contractResolver.ResolveContract(memberExpression.Expression.Type);
```
rynowak: Are we guaranteed that this will be a valid cast? Could you possibly construct a unit test case where this cast will fail?

src/Microsoft.AspNetCore.JsonPatch/JsonPatchDocumentOfT.cs
```diff
@@ -93,7 +93,7 @@ public JsonPatchDocument<TModel> Add<TProp>(Expression<Func<TModel, TProp>> path
 
             Operations.Add(new Operation<TModel>(
                 "add",
-                ExpressionHelpers.GetPath(path).ToLowerInvariant() + "/" + position,
+                ExpressionHelpers.GetPath(path, ContractResolver).ToLowerInvariant() + "/" + position,
```
rynowak: Are these 'helpers' used in more places than this class?

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPatchDocumentJsonPropertyAttributeTest.cs
```diff
@@ -4,6 +4,8 @@
 using Newtonsoft.Json;
 using System.Linq;
 using Xunit;
+using Newtonsoft.Json.Serialization;
+using System.Reflection;
```
rynowak: Sort

src/Microsoft.AspNetCore.JsonPatch/JsonPatchDocumentOfT.cs
kichalla: No, its being used only in this class as of now. Are you suggesting a name change or make it stateful type by supplying the resolver?

rynowak: I'm suggesting that we get rid of the 'helpers' because it's really just the core functionality of this class by another name.

kichalla:
e8452821b9a7689853e36c9a10ad4b8c6281817c

PR #55, opened 2017-01-24
[Fixes #48] Regression: List add must support adding items where index is same as count of elements

### LINKED ISSUE #48: Simple add on Lists failing in 1.1.0+
Working on 1.0.0, but failing on 1.1.0 and 1.2.0-preview1-22759

Here I'm attempting to add a new entry on List named Mylist which already has two entries (at indexes zero and one)

`[{"op":"add","path":"/mylist/2","value":{"myname":"foo","age":33}}]`

Response:
JsonPatchException: The index value provided by path segment '2' is out of bounds of the array size.

I'm guessing it's incorrectly looking at the length of the list prior to adding. With a dash at the end instead of the new index, this **will** work:

`[{"op":"add","path":"/mylist/-","value":{"myname":"foo","age":33}}]`

kichalla:
…x is same as count of elements

NOTE: this fix is for 1.1.1 branch

### DIFF
--- a/src/Microsoft.AspNetCore.JsonPatch/Internal/ListAdapter.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/Internal/ListAdapter.cs
@@ -27,7 +27,7 @@ namespace Microsoft.AspNetCore.JsonPatch.Internal
             }
 
             PositionInfo positionInfo;
-            if (!TryGetPositionInfo(list, segment, out positionInfo, out errorMessage))
+            if (!TryGetPositionInfo(list, segment, OperationType.Add, out positionInfo, out errorMessage))
             {
                 return false;
             }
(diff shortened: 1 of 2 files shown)


jbagga:
Any reason why [`ListAdapterTests`](https://github.com/aspnet/JsonPatch/blob/f314d7fb156d95a814868b62f31b3faeba6401d5/test/Microsoft.AspNetCore.JsonPatch.Test/ListAdapterTest.cs) do not have coverage for any other methods except `TryAdd` or `TryReplace`? Are they covered in another set of tests?

src/Microsoft.AspNetCore.JsonPatch/Internal/ListAdapter.cs
```diff
@@ -261,6 +266,14 @@ private bool TryGetPositionInfo(IList list, string segment, out PositionInfo pos
                     errorMessage = null;
                     return true;
                 }
+                // As per JSON Patch spec, for Add operation the index value representing the number of elements is valid,
+                // where as for other operations like Remove, Replace, Move and Copy the target index MUST exist.
+                else if (position == list.Count && operationType == OperationType.Add)
+                {
+                    positionInfo = new PositionInfo(PositionType.EndOfList, -1);
```
rynowak: Will this muck anything else up? Since we've now *lost* the fact that the index was provided, and the index was the end of the list, not `-1`

kichalla: No this does not cause any issues because we consider value of the index only when when the PositionType is Index.

PR #63, opened 2017-02-07
[Fixes #61] Move must keep object reference

### LINKED ISSUE #61: Incorrect move operation behavior
When performing a `move` operation, target object is not moved, but it is removed from source and **a copy** is added to the destination.
There is no problem with value types, but there is one with reference types : reference was lost.

To handle move operation, `ObjectAdapter` gets the moved property, removes it from the source, then adds it to the destination. That's OK.

The issue is in `ListAdapter`, `PocoAdapter` and `ExpandoAdapter` (don't know why, but not in `DictionaryAdapter`), in `TryConvertValue` method where the value is copied regardless of its type.

If object's type is compatible with target property's type, the reference should be kept.

TFleury:


Eilon:
@kichalla I assigned to you regarding https://github.com/aspnet/JsonPatch/issues/61.

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -362,6 +363,22 @@ public void Copy(Operation operation, object objectToApplyTo)
...
+                    }
+                    else
+                    {
+                        var error = CreateOperationFailedError(objectToApplyTo, operation.path, operation, Resources.FormatCannotCopyProperty(operation.from));
+                        ErrorReporter(error);
+                        return;
+                    }
+                }
+
                 Add(operation.path,
```
kichalla: Since Add operation should either do deep copy or maintain reference based on context in which its used, we could add a flag to Add operations of adapters to either maintain reference or not. The reference check could also be moved to `ConversionResultProvider`.

test/Microsoft.AspNetCore.JsonPatch.Test/NestedObjectTests.cs
```diff
@@ -1700,6 +1700,80 @@ public void CopyToEndOfListWithSerialization()
         }
 
         [Fact]
+        public void CopyDeepCloneObject()
```
kichalla: nit: Copy_DeepClonesObject

test/Microsoft.AspNetCore.JsonPatch.Test/NestedObjectTests.cs
```diff
@@ -1700,6 +1700,80 @@ public void CopyToEndOfListWithSerialization()
...
+            var patchDoc = new JsonPatchDocument<SimpleDTOWithNestedDTO>();
+            patchDoc.Copy<SimpleDTO>(o => o.InheritedDTO, o => o.SimpleDTO);
+
+            // Act
+            patchDoc.ApplyTo(doc);
+
+            // Assert
+            Assert.Equal("C", doc.SimpleDTO.StringProperty);
+            Assert.Equal("D", doc.SimpleDTO.AnotherStringProperty);
+            Assert.Equal("C", doc.InheritedDTO.StringProperty);
```
kichalla: Assert.NotSame(doc.SimpleDTO.StringProperty, doc.InheritedDTO.StringProperty);

test/Microsoft.AspNetCore.JsonPatch.Test/NestedObjectTests.cs
```diff
@@ -1753,6 +1827,77 @@ public void MoveWithSerialization()
         }
 
         [Fact]
+        public void MoveReference()
```
kichalla: nit: Move_KeepsObjectReference

test/Microsoft.AspNetCore.JsonPatch.Test/NestedObjectTests.cs
```diff
@@ -1802,6 +1947,71 @@ public void MoveInListWithSerialization()
         }
 
         [Fact]
+        public void MoveReferenceInList()
```
kichalla: nit: Move_KeepsObjectReferenceInList

test/Microsoft.AspNetCore.JsonPatch.Test/NestedObjectTests.cs
```diff
@@ -1802,6 +1947,71 @@ public void MoveInListWithSerialization()
...
+
+            // Assert
+            Assert.Equal(new List<SimpleDTO>() { sDto2, sDto1, sDto3 }, doc.SimpleDTOList);
+            Assert.Equal(2, doc.SimpleDTOList[0].IntegerValue);
+            Assert.Equal(1, doc.SimpleDTOList[1].IntegerValue);
+            Assert.Same(sDto2, doc.SimpleDTOList[0]);
+            Assert.Same(sDto1, doc.SimpleDTOList[1]);
+        }
+
+        [Fact]
```
kichalla: We would need tests for the other adapters which are impacted by the changes here: Poco and Expando

TFleury:
@kichalla, I don't think passing a flag requesting a deep copy to the `Add` was the best solution (but it was my first idea).
`Copy` method must be responsible of making the deep copy of the original value. Doing so, we don't introduce any breaking change to `IAdapter` interface.

But my first solution was a bit ugly, and serialization/deserialization was made 2 times for a copy.

As requested, I moved type compatibility checks to `ConversionResultProvider`. I changed `ConvertTo` behavior making it trying to keep reference when possible, and added `CopyTo` method when we need to enforce deep copy.

Can you review latest commits and let me know what you think about it ?

kichalla:
42d0d40b36df498f6367326810852ac546c81352

Thanks! @TFleury

Issue #80, opened 2017-05-18
JsonPatchException: Invalid JsonPatch operation thrown on model binding 


hjaltijon:
When the operation is invalid, an exception will be thrown on model binding.
    {
      "value": "someValue",
      "path": "/validPath",
      "op": "invalidOperationName"
    }

Why is that? I would have thought that the patch document would become null, with an error message in the modelstate, just like when the json body is invalid? 
I dont want to return a 500 error to the caller when the operation is invalid...

Is this the intended behavior? If so, is there a workaround?

rynowak:
An exception/500 here is definitely not intended.

Is your action method getting called? Can you include the call stack that's throwing?

hjaltijon:
my action method looks like this:

```
[HttpPatch("users/{username}")]
[Authorize]
public async Task<IActionResult> UpsertUser([FromRoute] string username,
            [FromBody] JsonPatchDocument<UserForUpdateDTO> patchDoc)
{
      if (patchDoc == null) return BadRequest();
      ...
```

the code inside the action method will not be excecuted if the operation name is not valid,
for example with this as the request body:
```
[
    {
      "value": "someValue",
      "path": "/somePath",
      "op": "invalidOperation"
    }
]
```
an exception would be thrown.


But with this as the request body:
```
[
    {
      ThisIsNotValidJson :)
    }
]
```
the code inside the action method will excecute normally and the caller will get a 400 Bad Request response.


Here is the call stack:
```
Microsoft.AspNetCore.JsonPatch.Exceptions.JsonPatchException: Invalid JsonPatch operation 'invalidOperationName'.
   at Microsoft.AspNetCore.JsonPatch.Operations.OperationBase.get_OperationType()
   at Microsoft.Extensions.Internal.PropertyHelper.CallNullSafePropertyGetter[TDeclaringType,TValue](Func`2 getter, Object target)
   at Microsoft.AspNetCore.Mvc.Internal.DefaultComplexObjectValidationStrategy.Enumerator.MoveNext()
   at Microsoft.AspNetCore.Mvc.ModelBinding.Validation.ValidationVisitor.VisitChildren(IValidationStrategy strategy)
   at Microsoft.AspNetCore.Mvc.ModelBinding.Validation.ValidationVisitor.VisitComplexType()
   at Microsoft.AspNetCore.Mvc.ModelBinding.Validation.ValidationVisitor.Visit(ModelMetadata metadata, String key, Object model)
[...]

PR #87, opened 2017-06-20
Refactor invalid operation exception handling

jbagga:
Addresses #80 

WIP for the MVC change.

cc @kichalla @rynowak

### DIFF
--- a/test/Microsoft.AspNetCore.JsonPatch.Test/OperationBaseTests.cs
+++ b/test/Microsoft.AspNetCore.JsonPatch.Test/OperationBaseTests.cs
@@ -8,12 +8,12 @@ namespace Microsoft.AspNetCore.JsonPatch.Operations
     public class OperationBaseTests
     {
         [Theory]
-        [InlineData("add", OperationType.Add)]
-        [InlineData("copy", OperationType.Copy)]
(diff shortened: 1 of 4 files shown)


jbagga:
The existing tests in JsonPatchDocumentTest still work as before even after the refactor. We don't expect any change in behavior.

jbagga:
@rynowak Added tests

test/Microsoft.AspNetCore.JsonPatch.Test/OperationBaseTests.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    public class OperationBaseTests
+    {
+        [Theory]
+        [InlineData("add", OperationType.Add)]
+        [InlineData("copy", OperationType.Copy)]
+        [InlineData("move", OperationType.Move)]
+        [InlineData("remove", OperationType.Remove)]
+        [InlineData("replace", OperationType.Replace)]
+        [InlineData("test", OperationType.Test)]
```
rynowak: Include some variations of casing since this is case-insensitive.

jbagga:
I'll merge this after ensuring it doesn't break MVC. Thanks!

Issue #38, opened 2016-09-27
Support for patching dynamic types should go beyond ExpandoObject


kichalla:
Currently we handle patching of dynamic types like ExpandoObject, but there are also other dynamic types like `DynamicObject` which should be supported. 

All dynamic types are supposed to implement `IDynamicMetaObjectProvider` which we can look for while trying to support this behavior.

@rynowak @dougbu

Eilon:
Putting on backlog, awaiting customer feedback.

TAGC:
Has this been resolved or is anywhere near close to being resolved? .NET Core Preview 2 SDK seems to treat downgrades as errors rather than warnings, and this is the _only_ dependency I'm forced to downgrade (to v1.0.0) in order to workaround the problems discussed in #59.

jbagga:
Not yet. This is planned for 2.1

TAGC:
After updating the Microsoft.* NuGet packages to the recently released 2.0.0 versions, I'm no longer able to use the downgraded 1.0.0 version of `Microsoft.AspNetCore.JsonPatch`. I get an exception within the `Startup.Configure()` method of my web service:

```
Autofac.Core.DependencyResolutionException occurred
  HResult=0x80131500
  Message=An error occurred during the activation of a particular registration. See the inner exception for details. Registration: Activator = MvcRouteHandler (ReflectionActivator), Services = [Microsoft.AspNetCore.Mvc.Internal.MvcRouteHandler], Lifetime = Autofac.Core.Lifetime.RootScopeLifetime, Sharing = Shared, Ownership = OwnedByLifetimeScope
  Source=Autofac
  StackTrace:
   at Autofac.Core.Resolving.InstanceLookup.Activate(IEnumerable`1 parameters)
   at Autofac.Core.Lifetime.LifetimeScope.GetOrCreateAndShare(Guid id, Func`1 creator)
   at Autofac.Core.Resolving.InstanceLookup.Execute()
   at Autofac.Core.Resolving.ResolveOperation.GetOrCreateInstance(ISharingLifetimeScope currentOperationScope, IComponentRegistration registration, IEnumerable`1 parameters)
   at Autofac.Core.Resolving.ResolveOperation.Execute(IComponentRegistration registration, IEnumerable`1 parameters)
   at Autofac.ResolutionExtensions.TryResolveService(IComponentContext context, Service service, IEnumerable`1 parameters, Object& instance)
[...]

PR #99, opened 2017-08-07
Support DynamicObject types

jbagga:
Addresses #38 

cc @kichalla @rynowak


src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -97,15 +97,15 @@ public class ExpandoObjectAdapter : IAdapter
...
+            var dynamicObject = target as DynamicObject;
+            if (dynamicObject == null)
             {
                 errorMessage = null;
                 nextTarget = null;
                 return false;
             }
 
-            var dictionary = (IDictionary<string, object>)expandoObject;
+            var dictionary = (IDictionary<string, object>)dynamicObject;
```
rynowak: This isn't a good assumption. We need to use `IDynamicMetaObjectProvider` directly.

jbagga:
@rynowak Added a new adapter instead of replacing `ExpandoObjectAdapter`.

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -0,0 +1,221 @@
...
+    public class DynamicObjectAdapter : IAdapter
+    {
+        public bool TryAdd(
+            object target,
+            string segment,
+            IContractResolver contractResolver,
+            object value,
+            out string errorMessage)
+        {
+            if(!TrySetDynamicObjectProperty(target, contractResolver, segment, value))
```
rynowak: spacing

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -0,0 +1,221 @@
...
+            out object nextTarget,
+            out string errorMessage)
+        {
+            var dynamicObject = target as IDynamicMetaObjectProvider;
+            if (dynamicObject == null)
+            {
+                nextTarget = null;
+                errorMessage = null;
+                return false;
+            }
```
rynowak: Under what conditions can this happen? Should we be checking this in all of these methods and returning a good error message?

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -0,0 +1,221 @@
...
+            object target,
+            IContractResolver contractResolver,
+            string segment,
+            out object property)
+        {
+            var jsonDynamicContract = contractResolver.ResolveContract(target.GetType()) as JsonDynamicContract;
+
+            var propertyName = jsonDynamicContract?.PropertyNameResolver(segment);
+
+            var binder = CSharp.RuntimeBinder.Binder.GetMember(CSharpBinderFlags.None,
```
rynowak: newline after `(`

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -0,0 +1,221 @@
...
+                propertyName,
+                target.GetType(),
+                new List<CSharpArgumentInfo>{
```
rynowak: https://en.wikipedia.org/wiki/Indent_style#Allman_style

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -0,0 +1,221 @@
...
+            }
+        }
+
+        private bool TryGetDynamicObjectProperty(
+            object target,
+            IContractResolver contractResolver,
+            string segment,
+            out object property)
+        {
+            var jsonDynamicContract = contractResolver.ResolveContract(target.GetType()) as JsonDynamicContract;
```
rynowak: Under what conditions do we expect this cast to fail?

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -0,0 +1,221 @@
...
+
+            var propertyName = jsonDynamicContract?.PropertyNameResolver(segment);
```
rynowak: You'll get an ArgumentNullException later at line 156. If you are going to be defensive, make sure you are defending from the right things

jbagga:
@rynowak Updated

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -146,16 +130,19 @@ public class DynamicObjectAdapter : IAdapter
...
         {
             var jsonDynamicContract = contractResolver.ResolveContract(target.GetType()) as JsonDynamicContract;
 
             var propertyName = jsonDynamicContract?.PropertyNameResolver(segment);
 
-            var binder = CSharp.RuntimeBinder.Binder.GetMember(CSharpBinderFlags.None,
-                propertyName,
+            var binder = CSharp.RuntimeBinder.Binder.GetMember(
+                CSharpBinderFlags.None,
+                propertyName ?? segment,
```
rynowak: Why is this right?

jbagga: @rynowak  If `segment` is null it will throw [here](https://github.com/aspnet/JsonPatch/blob/dev/src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs#L129) before it gets to the adapter.

If the `contract` or the `propertName` cannot be resolved, segment may still be a valid path for the patch document. Is that risky?

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        private bool TryGetDynamicObjectProperty(
+            object target,
+            IContractResolver contractResolver,
+            string segment,
+            out object property,
+            out string errorMessage)
+        {
+            var jsonDynamicContract = contractResolver.ResolveContract(target.GetType()) as JsonDynamicContract;
```
rynowak: When can this cast fail? I think you're being defensive here with something that can't happen

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -0,0 +1,215 @@
...
+
+            var propertyName = jsonDynamicContract?.PropertyNameResolver(segment);
+
+            var binder = CSharp.RuntimeBinder.Binder.GetMember(
```
rynowak: Add a .using if possible. If there's a conflict then use `using CSharpBinder = ...`

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -0,0 +1,215 @@
...
+                errorMessage = null;
+                return true;
+            }
+        }
+
+        private bool TryGetDynamicObjectProperty(
+            object target,
+            IContractResolver contractResolver,
+            string segment,
+            out object property,
```
rynowak: This isn't the right name, it's the property's value not the property itself

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -0,0 +1,215 @@
...
+            var propertyName = jsonDynamicContract?.PropertyNameResolver(segment);
+
+            var binder = CSharp.RuntimeBinder.Binder.GetMember(
+                CSharpBinderFlags.None,
+                propertyName ?? segment,
+                target.GetType(),
+                new List<CSharpArgumentInfo>
+                {
+                    CSharpArgumentInfo.Create(CSharpArgumentInfoFlags.None, null),
+                    CSharpArgumentInfo.Create(CSharpArgumentInfoFlags.None, null)
```
rynowak: Is this right? I would expect the getter to have only one argument (`this`).

test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/DynamicObjectAdapterTests.cs
```diff
@@ -0,0 +1,1215 @@
...
+            };
+
+            // create patch
+            JsonPatchDocument patchDoc = new JsonPatchDocument();
+            patchDoc.Add("DynamicProperty/StringProperty", "B");
+
+            var serialized = JsonConvert.SerializeObject(patchDoc);
+            var deserialized = JsonConvert.DeserializeObject<JsonPatchDocument>(serialized);
+
+            deserialized.ApplyTo(doc);
```
rynowak: Why do you need this? I think that you should be able to use `patchDoc` directly

test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/DynamicObjectAdapterTests.cs
```diff
@@ -0,0 +1,1215 @@
...
+            deserialized.ApplyTo(obj);
+
+            Assert.Equal(1, obj.NewInt);
+            Assert.Equal(1, obj.Test);
+        }
+
+        [Fact]
+        public void AddNewPropertyToDynamicOjectInTypedObject()
+        {
+            var doc = new NestedDTO()
```
rynowak: avoid abbreviations like this

test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/DynamicObjectAdapterTests.cs
```diff
@@ -0,0 +1,1215 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using Microsoft.AspNetCore.JsonPatch.Exceptions;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Microsoft.AspNetCore.JsonPatch.Test.Dynamic
```
rynowak: This isn't really the right namespace.

test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/DynamicObjectAdapterTests.cs
```diff
@@ -0,0 +1,1215 @@
...
+{
+    public class DynamicObjectAdapterTests
```
rynowak: If these are unit tests then there should be calling methods on the adapter directly.

If these are integration tests, we should organize them as such.

rynowak: https://github.com/aspnet/Home/wiki/Engineering-guidelines#unit-test-structure

test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/DynamicObjectAdapterTests.cs
```diff
@@ -0,0 +1,1215 @@
...
+            var serialized = JsonConvert.SerializeObject(patchDoc);
+            var deserialized = JsonConvert.DeserializeObject<JsonPatchDocument>(serialized);
+
+            deserialized.ApplyTo(doc);
+
+            Assert.Equal(new List<int>() { 4, 5, 6 }, doc.IntegerList);
+        }
+
+        [Fact]
+        public void ReplaceAtEndOfList()
```
rynowak: This is wayyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy overtesting for dynamic. Most of these really feel like integration tests.

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
rynowak: Yeah but **under what conditions** would any of these things happen?

test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/DynamicObjectAdapterTests.cs
jbagga: I tried to be consistent with https://github.com/aspnet/JsonPatch/blob/dev/test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/AddOperationTests.cs and other tests [here](https://github.com/aspnet/JsonPatch/tree/dev/test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic). Should I move my tests to the existing test classes? They are organized by `OperationType`, not by `Adapters`.

test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/DynamicObjectAdapterTests.cs
jbagga: So rename [this](https://github.com/aspnet/JsonPatch/blob/dev/test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/NestedDTO.cs)?

test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/DynamicObjectAdapterTests.cs
jbagga: Yes that's how I wrote them but again for consistency with existing tests I changed them to use `deserialized`. I'll change them to use `patchDoc` instead

test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/DynamicObjectAdapterTests.cs
rynowak: I'm sending one of my minions to help you sort out this mess later today. Don't fret, help is on the way

test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/DynamicObjectAdapterTests.cs
rynowak: Neither `doc` nor `NestedDTO` are great names

test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic/DynamicObjectAdapterTests.cs
rynowak: Yeah, don't copy bad stuff 😆

jbagga:
@rynowak Added unit tests

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        private bool TryGetDynamicObjectProperty(
+            object target,
+            IContractResolver contractResolver,
+            string segment,
+            out object value,
+            out string errorMessage)
+        {
+            var jsonDynamicContract = contractResolver.ResolveContract(target.GetType()) as JsonDynamicContract;
```
rynowak: Please stop by so we can discuss this in person. I've commented on this about 4 times and haven't yet heard and answer.

test/Microsoft.AspNetCore.JsonPatch.Test/DynamicObjectAdapterTest.cs
```diff
@@ -0,0 +1,237 @@
...
+            var status = adapter.TryTraverse(target.NestedObject, segment, resolver, out object nextTarget, out string errorMessage);
+
+            // Assert
+            Assert.False(status);
+            Assert.Equal($"The target location specified by path segment '{segment}' was not found.", errorMessage);
+        }
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData("new")]
```
rynowak: One small comment here. 

Resist the urge to **prove** things using multiple values.  One of these cases is as good as any other and there's no reason to have them both.

Save theory tests for cases where you have a complicated algorithm to verify. The more experience I gain, the less often I use them.

test/Microsoft.AspNetCore.JsonPatch.Test/DynamicObjectAdapterTest.cs
```diff
@@ -0,0 +1,237 @@
...
+            var segment = "NewProperty";
+            var resolver = new DefaultContractResolver();
+
+            // Act
+            var removeStatus = adapter.TryRemove(target, segment, resolver, out string removeErrorMessage);
+
+            // Assert
+            Assert.False(removeStatus);
+            Assert.Equal($"The target location specified by path segment '{segment}' was not found.", removeErrorMessage);
+        }
```
rynowak: These are good 👍

jbagga:
@rynowak Updated

PR #106, opened 2017-08-30
Combine ExpandoObjectAdapter and DictionaryAdapter

jbagga:
Refactoring some code to combine the two (similar) adapters, ensuring both types use the `JsonDictionaryContract` and resolve the dictionary keys.

May need more tests. This is just to get feedback on some generic type conversion. See my comment below.

cc @rynowak


src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
```diff
@@ -133,11 +138,50 @@ private object ConvertValue(IDictionary<string, object> dictionary, string key,
...
-                        return conversionResult.ConvertedInstance;
+                        var convertedInstance = conversionResult.ConvertedInstance;
+                        return TryParse<TValue>(convertedInstance);
                     }
                 }
             }
             return newValue;
         }
+
+        private static T TryParse<T>(string value)
```
jbagga: @rynowak Can we talk about what's the best way to either combine the two `TryParse` methods or if there is a better way to convert keys and values to `TKey` and `TValue`?

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
```diff
@@ -18,10 +18,12 @@ public class ExpandoObjectAdapter : IAdapter
...
             var contract = (JsonDictionaryContract)contractResolver.ResolveContract(target.GetType());
             var key = contract.DictionaryKeyResolver(segment);
-            var dictionary = (IDictionary<string, object>)target;
+            var dictionary = (IDictionary<TKey, TValue>)target;
 
             // As per JsonPatch spec, if a key already exists, adding should replace the existing value
-            dictionary[key] = ConvertValue(dictionary, key, value);
+            var convertedKey = TryParse<TKey>(key);
+            var convertedValue = TryParse<TValue>(value);
+            dictionary[convertedKey] = ConvertValue(dictionary, convertedKey, convertedValue);
```
kichalla: Instead of TryParse methods, could you use `ConversionResultProvider.ConvertTo` for `TKey` (this could also solve the issue of non-string keys in dictionary) and use `ConvertValue` for `TValue` (you need not use TryParse for value as you are anyway going to convert the value)?

jbagga:
@kichalla @rynowak Updated

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
```diff
@@ -18,10 +18,11 @@ public class ExpandoObjectAdapter : IAdapter
         {
             var contract = (JsonDictionaryContract)contractResolver.ResolveContract(target.GetType());
             var key = contract.DictionaryKeyResolver(segment);
-            var dictionary = (IDictionary<string, object>)target;
+            var dictionary = (IDictionary<TKey, TValue>)target;
 
             // As per JsonPatch spec, if a key already exists, adding should replace the existing value
-            dictionary[key] = ConvertValue(dictionary, key, value);
+            var convertedKey = (TKey)ConversionResultProvider.ConvertTo(key, typeof(TKey)).ConvertedInstance;
```
rynowak: What if this conversion fails? Casting a null value to `int` for instance will throw an exception

src/Microsoft.AspNetCore.JsonPatch/Internal/ObjectVisitor.cs
```diff
@@ -49,21 +51,20 @@ public bool TryVisit(ref object target, out IAdapter adapter, out string errorMe
...
-                return new DynamicObjectAdapter();
+                return new ListAdapter();
             }
-            else if (targetObject is IDictionary)
+            else if (jsonContract is JsonDictionaryContract jsonDictionaryContract)
             {
-                return new DictionaryAdapter();
+                var type = typeof(DictionaryAdapter<,>).MakeGenericType(jsonDictionaryContract.DictionaryKeyType, jsonDictionaryContract.DictionaryValueType);
+                return (IAdapter)Activator.CreateInstance(type);
             }
```
rynowak: We may still need to handle the case where an object is an `IDictionary` but not an `IDictionary<>`

test/Microsoft.AspNetCore.JsonPatch.Test/DictionaryAdapterTest.cs
```diff
@@ -18,11 +17,11 @@ public void Add_KeyWhichAlreadyExists_ReplacesExistingValue()
             var nameKey = "Name";
             var dictionary = new Dictionary<string, object>(StringComparer.Ordinal);
             dictionary[nameKey] = "Mike";
-            var dictionaryAdapter = new DictionaryAdapter();
-            var resolver = new Mock<IContractResolver>(MockBehavior.Strict);
+            var dictionaryAdapter = new DictionaryAdapter<string, object>();
```
rynowak: We need some tests for things besides `<string, object>`

test/Microsoft.AspNetCore.JsonPatch.Test/DictionaryAdapterTest.cs
```diff
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
```
rynowak: Do the expando object tests need an update?

jbagga: There are no unit tests for ExpandoObjectAdapter. Only integration tests, which pass using the new adapter

test/Microsoft.AspNetCore.JsonPatch.Test/DictionaryAdapterTest.cs
jbagga: Working on adding these

jbagga:
@rynowak Changed key and value conversion code

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
```diff
@@ -18,11 +17,22 @@ public class ExpandoObjectAdapter : IAdapter
         {
             var contract = (JsonDictionaryContract)contractResolver.ResolveContract(target.GetType());
             var key = contract.DictionaryKeyResolver(segment);
-            var dictionary = (IDictionary<string, object>)target;
+            var dictionary = (IDictionary<TKey, TValue>)target;
 
             // As per JsonPatch spec, if a key already exists, adding should replace the existing value
-            dictionary[key] = ConvertValue(dictionary, key, value);
+            var canConvertKey = TryConvertKey(key, out var convertedKey, out errorMessage);
+            if (!canConvertKey)
```
rynowak: You can just merge this into the `if` if you want it to be more terse.

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
```diff
@@ -124,20 +160,38 @@ public class ExpandoObjectAdapter : IAdapter
...
+            var conversionResult = ConversionResultProvider.ConvertTo(value, typeof(TValue));
+            if (conversionResult.CanBeConverted)
+            {
+                errorMessage = null;
+                convertedValue = (TValue)conversionResult.ConvertedInstance;
+                return true;
+            }
+            else
+            {
+                errorMessage = Resources.FormatInvalidValueForProperty(value);
```
rynowak: This is a good message `The value '{0}' is invalid for target location.`

src/Microsoft.AspNetCore.JsonPatch/Internal/ObjectVisitor.cs
rynowak: We've decided not to do anything about this.

test/Microsoft.AspNetCore.JsonPatch.Test/DictionaryAdapterTest.cs
rynowak: lel ok

rynowak:
This is looking pretty good now. Just waiting on a new error message and some tests.

jbagga:
@rynowak Updated with tests

Issue #98, opened 2017-08-06
// root path for List<> type


rosspi:
I am getting a path string to be generated from an expression to the root object that looks like this:
//
That is, x => x is turning into '//' when it should just be '/'.
This is just the case when the generic type of the document is a list.

When applying the apply function(s) it seems to be happy with the double slash, but I don't think this should be the case.

Eilon:
@jbagga can you take a look and see what's going on here?

jbagga:
@rosspi Can you share a sample/code snippet, and elaborate on how the actual and expected output differ? I can then investigate.

rosspi:
It happens when the generic type is a list, and you're adding an item to the object itself (when using an expression like this: x => x).
If however, the expression points at something deeper than the object itself, then the resultant json path is ok.

```
class MyClass
{
        
}
class Program
{
    static void Main(string[] args)
    {
        var blah = new JsonPatchDocument<List<MyClass>>();
        blah.Add(x => x, new MyClass()); 

        Console.WriteLine(blah.Operations[0].path);
    }
}
```

PR #107, opened 2017-09-05
Refactor GetPath in JsonPatchDocumentOfT

jbagga:
Addresses #98 

cc @kichalla

### DIFF
--- a/src/Microsoft.AspNetCore.JsonPatch/JsonPatchDocumentOfT.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/JsonPatchDocumentOfT.cs
@@ -58,7 +58,7 @@ namespace Microsoft.AspNetCore.JsonPatch
 
             Operations.Add(new Operation<TModel>(
                 "add",
-                GetPath(path),
+                GetPath(path, null),
                 from: null,
                 value: value));
 
@@ -85,7 +85,7 @@ namespace Microsoft.AspNetCore.JsonPatch
 
             Operations.Add(new Operation<TModel>(
                 "add",
-                GetPath(path) + "/" + position,
+                GetPath(path, position.ToString()),
                 from: null,
                 value: value));
 
@@ -108,7 +108,7 @@ namespace Microsoft.AspNetCore.JsonPatch
 
             Operations.Add(new Operation<TModel>(
                 "add",
-                GetPath(path) + "/-",
+                GetPath(path, "-"),
                 from: null,
                 value: value));
 
@@ -128,7 +128,7 @@ namespace Microsoft.AspNetCore.JsonPatch
                 throw new ArgumentNullException(nameof(path));
             }
 
-            Operations.Add(new Operation<TModel>("remove", GetPath(path), from: null));
+            Operations.Add(new Operation<TModel>("remove", GetPath(path, null), from: null));
 
             return this;
         }
@@ -149,7 +149,7 @@ namespace Microsoft.AspNetCore.JsonPatch
 
             Operations.Add(new Operation<TModel>(
                 "remove",
-                GetPath(path) + "/" + position,
+                GetPath(path, position.ToString()),
                 from: null));
 
             return this;
@@ -170,7 +170,7 @@ namespace Microsoft.AspNetCore.JsonPatch
 
             Operations.Add(new Operation<TModel>(
                 "remove",
-                GetPath(path) + "/-",
+                GetPath(path, "-"),
                 from: null));
 
             return this;
@@ -192,7 +192,7 @@ namespace Microsoft.AspNetCore.JsonPatch
 
             Operations.Add(new Operation<TModel>(
                 "replace",
-                GetPath(path),
+                GetPath(path, null),
                 from: null,
                 value: value));
 
@@ -217,7 +217,7 @@ namespace Microsoft.AspNetCore.JsonPatch
 
             Operations.Add(new Operation<TModel>(
                 "replace",
-                GetPath(path) + "/" + position,
+                GetPath(path, position.ToString()),
                 from: null,
                 value: value));
 
@@ -240,7 +240,7 @@ namespace Microsoft.AspNetCore.JsonPatch
 
             Operations.Add(new Operation<TModel>(
                 "replace",
-                GetPath(path) + "/-",
+                GetPath(path, "-"),
                 from: null,
                 value: value));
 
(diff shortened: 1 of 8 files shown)


test/Microsoft.AspNetCore.JsonPatch.Test/JsonPatchDocumentJsonPropertyAttributeTest.cs
```diff
@@ -12,6 +13,40 @@ namespace Microsoft.AspNetCore.JsonPatch
     public class JsonPatchDocumentJsonPropertyAttributeTest
     {
         [Fact]
+        public void Add_ToRoot_OfListOfObjects()
+        {
+            var patchDoc = new JsonPatchDocument<List<JsonPropertyObject>>();
+            patchDoc.Add(p => p, new JsonPropertyObject());
+
+            var serialized = JsonConvert.SerializeObject(patchDoc);
+            // serialized value should have "AnotherName" as path
```
kichalla: this comment is not valid here...probably a copy-paste error

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPatchDocumentJsonPropertyAttributeTest.cs
```diff
@@ -12,6 +13,40 @@ namespace Microsoft.AspNetCore.JsonPatch
...
+            // get path
+            var pathToCheck = deserialized.Operations.First().path;
+            Assert.Equal("/-", pathToCheck);
+        }
+
+        [Fact]
+        public void Add_ToRoot_OfAnObject()
+        {
+            var patchDoc = new JsonPatchDocument<JsonPropertyObject>();
+            patchDoc.Add(p => p, new JsonPropertyObject());
```
kichalla: This test doesn't make much sense here. (Adding to a POCO object itself). Also can Path be empty? i suppose that's an error case.

src/Microsoft.AspNetCore.JsonPatch/JsonPatchDocumentOfT.cs
```diff
@@ -717,7 +717,13 @@ IList<Operation> IJsonPatchDocument.GetOperations()
...
         private string GetPath<TProp>(Expression<Func<TModel, TProp>> expr)
         {
-            return "/" + GetPath(expr.Body, true).ToLowerInvariant();
+            var path = GetPath(expr.Body, true);
+            if (!String.IsNullOrEmpty(path))
+            {
+                path = "/" + path;
+            }
+
+            return path.ToLowerInvariant();
```
rynowak: Should this change just go into the `case ExpressionType.Parameter` block?

Without some examples I'm a little skeptical that this fixes all of the cases we care about.

jbagga: Can we discuss this? I am not sure how I can preemptively remove the extra slash in the `case ExpressionType.Parameter` block?

jbagga:
@kichalla Added unit tests for GetPath

jbagga:
@rynowak @kichalla Updated

src/Microsoft.AspNetCore.JsonPatch/JsonPatchDocumentOfT.cs
```diff
@@ -776,33 +772,24 @@ private string GetPropertyNameFromMemberExpression(MemberExpression memberExpres
...
+                || expressionType == ExpressionType.Call
+                || expressionType == ExpressionType.Convert
+                || expressionType == ExpressionType.MemberAccess);
+
         }
 
-        private static string GetIndexerInvocation(Expression expression)
+        // Evaluates the value of the key or index which may be an int or a string, 
+        // or some other expression type.
+        // The expression is converted to executable code and the resulting delegate is returned as a string.
```
rynowak: `The expression is converted to a delegate and the result of executing the delegate is returned as a string`

src/Microsoft.AspNetCore.JsonPatch/JsonPatchDocumentOfT.cs
```diff
@@ -705,61 +705,57 @@ IList<Operation> IJsonPatchDocument.GetOperations()
...
+                {
+                    return path.ToLowerInvariant();
+                }
+            }
+
+            return "/" + path.ToLowerInvariant();
         }
 
-        private string GetPath(Expression expr, bool firstTime)
+        private List<string> GetPath(Expression expr, bool firstTime)
```
rynowak: I think the `first time` parameter can be removed

rynowak: It wouldn't hurt to rename this to something like `GetPathSegments` its usage isn't really interchangable with `GetPath` that returns a string

test/Microsoft.AspNetCore.JsonPatch.Test/GetPathTest.cs
```diff
@@ -0,0 +1,111 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Microsoft.AspNetCore.JsonPatch
+{
+    public class GetPathTest
```
rynowak: This should be called `JsonPatchDocumentGetPathTest`.

src/Microsoft.AspNetCore.JsonPatch/JsonPatchDocumentOfT.cs
```diff
@@ -705,61 +705,57 @@ IList<Operation> IJsonPatchDocument.GetOperations()
...
+            var segments = GetPathSegments(expr.Body);
+            var path = String.Join("/", segments);
+            if (position != null)
+            {
+                path += "/" + position;
+                if (segments.Count == 0)
+                {
+                    return path.ToLowerInvariant();
+                }
+            }
```
rynowak: what about a case where `segments.Count == 0` and `position != null`? It looks like the value would be missing the leading slash.

rynowak: A good way to think about this is that `path = "/" + string.Join("/", segments)` will always generate the right thing even when segments is empty.

Then you just have to deal with appending the position.

jbagga: This case is covered. Adding tests where position is not null

jbagga:
@rynowak @kichalla Added non-null positions tests

PR #111, opened 2017-10-02
Improve conformance of replace operations to spec

TAGC:
Addresses #110

This change makes it so that "replace" operations will remove existing values at a path before setting the new value.

Submitting this PR partly because I have a colleague desperately waiting on a fix that's blocked by this issue, but mainly because this should count as 1 of my 4 for Hacktoberfest.

I've modified an existing unit test to expose the bug and confirm the fix. I've also confirmed these changes fix the issue I was having.

### DIFF
--- a/src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
@@ -94,6 +94,11 @@ namespace Microsoft.AspNetCore.JsonPatch.Internal
                 return false;
             }
 
+            if (!TryRemove(target, segment, contractResolver, out errorMessage))
+            {
+                return false;
+            }
+
             if (!TrySetDynamicObjectProperty(target, contractResolver, segment, convertedValue, out errorMessage))
             {
                 return false;
--- a/test/Microsoft.AspNetCore.JsonPatch.Test/WriteOnceDynamicTestObject.cs
+++ b/test/Microsoft.AspNetCore.JsonPatch.Test/WriteOnceDynamicTestObject.cs
@@ -1,0 +1,117 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Dynamic;
+
+namespace Microsoft.AspNetCore.JsonPatch.Internal
+{
+    /// <remarks>
+    /// <para>
+    /// This class is used specifically to test that JSON patch "replace" operations are functionally equivalent to
+    /// "add" and "remove" operations applied sequentially using the same path.
+    /// </para>
+    /// <para>
+    /// This is done by asserting that  no value exists for a particular key before setting its value. To replace the
+    /// value for a key, the key must first be removed, and then re-added with the new value.
+    /// </para>
+    /// <para>
+    /// See JsonPatch#110 for further details.
+    /// </para>
+    /// </remarks>
+    public class WriteOnceDynamicTestObject : DynamicObject
+    {
+        private Dictionary<string, object> _dictionary = new Dictionary<string, object>();
+
+        public object this[string key] { get => ((IDictionary<string, object>)_dictionary)[key]; set => SetValueForKey(key, value); }
+
+        public ICollection<string> Keys => ((IDictionary<string, object>)_dictionary).Keys;
+
+        public ICollection<object> Values => ((IDictionary<string, object>)_dictionary).Values;
+
+        public int Count => ((IDictionary<string, object>)_dictionary).Count;
+
+        public bool IsReadOnly => ((IDictionary<string, object>)_dictionary).IsReadOnly;
+
+        public void Add(string key, object value)
+        {
+            SetValueForKey(key, value);
+        }
+
+        public void Add(KeyValuePair<string, object> item)
+        {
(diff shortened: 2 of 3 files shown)


rynowak:
@jbagga - can you please take a look at this?

test/Microsoft.AspNetCore.JsonPatch.Test/WriteOnceDynamicTestObject.cs
```diff
@@ -0,0 +1,117 @@
...
+            if (value == null)
+            {
+                _dictionary.Remove(key);
+                return;
+            }
+
+            if (_dictionary.ContainsKey(key))
+            {
+                throw new ArgumentException($"Value for {key} already exists");
+            }
```
jbagga: Why is this an exception? Curious why the new value is not set? `_dictionary[binder.Name] = value`

https://msdn.microsoft.com/en-us/library/zyxt2e2h(v=vs.110).aspx

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -94,6 +94,11 @@ public class DynamicObjectAdapter : IAdapter
                 return false;
             }
 
+            if (!TryRemove(target, segment, contractResolver, out errorMessage))
+            {
+                return false;
+            }
+
```
jbagga: The `TrySetDynamicObjectProperty` method, either creates a new entry in the underlying dictionary in the `DynamicObject` class or sets the existing property to the new value (in effect replacing the existing property). It shouldn't be necessary to remove it.

test/Microsoft.AspNetCore.JsonPatch.Test/WriteOnceDynamicTestObject.cs
TAGC: The purpose of this class (and the reason it's called "WriteOnce") is to assert that dictionary values are not overwritten, at least not without being removed first. From the documentation I added at the top:

> This is done by asserting that  no value exists for a particular key before setting its value. To replace the value for a key, the key must first be removed, and then re-added with the new value.

I modelled it similarly to the logic I use in my own project, which is how I discovered the issue.

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
TAGC: Right, but the issue as I described in #110 is that simply replacing the value using `TrySetDynamicObjectProperty` means that "replace" operations are not distinguishable from "add" operations. From RFC 6902, the operation should act as a remove operation (using `TrySetDynamicObjectProperty` to set the property to `null`) followed immediately by an add operation.

This is how you had [it implemented originally](https://github.com/aspnet/JsonPatch/blob/patch/1.0.5/src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs#L725) (setting `null` before setting the replacement value), so I believe this is an unintended breaking change. The issue I'm having now concerns an operation that I'm fairly sure I've tried before and didn't give me problems when I was using JSON Patch v1.0.0.

test/Microsoft.AspNetCore.JsonPatch.Test/WriteOnceDynamicTestObject.cs
jbagga: This seems like a very specific implementation requirement. The current `Replace` operation supports the generic scenario.

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
jbagga: "replace" is different from "add" because when replacing, the new value needs to be the same type as the original value assigned to the key.

jbagga:
Closing this for now as the current implementation supports the generic scenario.

test/Microsoft.AspNetCore.JsonPatch.Test/WriteOnceDynamicTestObject.cs
TAGC: > This seems like a very specific implementation requirement. 

Admittedly yeah, there's few cases where the distinction would be noticeable.

Still:
 - it's what the spec dictates. There should be no difference between doing a replace operation, and doing a remove operation followed by an add operation. I've shown one case where the current implementation doesn't adhere to that. Possibly others have encountered more.
 - It's the way you had it implemented originally. Changing it now constitutes a breaking change. Technically that's fine since you're a whole major version ahead at this point, but it seems like a gratuitous one.
 - Importantly, this change should not break any existing functionality.

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
TAGC: Sure, but my PR would not affect those semantics. My changes don't affect the final value that gets set for the property, it just temporarily replaces it with null beforehand (signalling property removal).

TAGC:
Kind of disappointing but fair enough. Would it be okay (legally) to push a NuGet package with these changes to my company’s private MyGet feed?

Eilon:
The spec doesn't dictate how an implementation needs to implement the replace operation, rather just that the ultimate object should be in a particular state after the operation has completed, so I don't think there's an issue there.

But, so long as you abide by the license used by this project (Apache 2.0 license), you are free to make any changes you want to a copy of the code and publish to your own feed.

Issue #105, opened 2017-08-24
Reorganize tests and remove redundant integration tests


jbagga:
The tests are organized in a confusing structure and not as we usually organize them. There are a lot of redundant integration tests and some adapters lack unit tests.

~~1. [This](https://github.com/json-patch/json-patch-tests/blob/master/tests.json) test suite can be used to ensure we are spec compliant. It can either be used to generate test code or parsed and tested.~~
- [x] 2. [This](https://github.com/aspnet/JsonPatch/tree/dev/test/Microsoft.AspNetCore.JsonPatch.Test/Dynamic) folder has lots of integration tests for all operations (we should remove redundancy)
- [x] 3. Add integration tests for "Test" operation
- [x] 4. Clean up test object model classes - simplify by adding them to test classes that use them
- [x] 5. Ensure all test classes follow the Arrange, Act, Assert pattern; variable names are coherent/sensible (replace patchDoc with patchDocument,  obj with object/targetObject etc.), remove unnecessary serialization and formatting of error messages
- [x] 6. `ExpandoObjectAdapter` and `PocoAdapter` do not have unit tests

[This](https://github.com/aspnet/JsonPatch/pull/99) PR adds a new adapter called `DynamicObjectAdapter` with [unit tests](https://github.com/aspnet/JsonPatch/blob/dev/test/Microsoft.AspNetCore.JsonPatch.Test/DynamicObjectAdapterTest.cs) and integration tests in a folder called [`IntegrationTests`](https://github.com/aspnet/JsonPatch/tree/dev/test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests).

cc @rynowak

jbagga:
`ExpandoObjectAdapter` was combined with `DictionaryAdapter`, so now it has unit tests. See [this](https://github.com/aspnet/JsonPatch/pull/106)

`PocoAdapter` unit tests were added [here](https://github.com/aspnet/JsonPatch/pull/114)

jbagga:
PR #117 cover 2., 3. and 5.

jbagga:
Shelving 1. for now. See #121

PR #114, opened 2017-10-13
Add Test Operation

jbagga:
Addresses #1 

Working on adding more integration tests for `Test` as part of the clean up for #105 

cc @Eilon


src/Microsoft.AspNetCore.JsonPatch/Adapters/IObjectAdapter.cs
```diff
@@ -15,5 +15,6 @@ public interface IObjectAdapter
         void Move(Operation operation, object objectToApplyTo);
         void Remove(Operation operation, object objectToApplyTo);
         void Replace(Operation operation, object objectToApplyTo);
+        void Test(Operation operation, object objectToApplyTo);
```
rynowak: @eilon, this is a breaking change.

The non-breaking way to do this is to introduce a base-class or another interface and conditionally cast to that.

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -365,6 +366,37 @@ public void Copy(Operation operation, object objectToApplyTo)
...
+                throw new ArgumentNullException(nameof(objectToApplyTo));
+            }
+
+            var parsedPath = new ParsedPath(operation.path);
+            var visitor = new ObjectVisitor(parsedPath, ContractResolver);
+
+            var target = objectToApplyTo;
+            if (!visitor.TryVisit(ref target, out var adapter, out var errorMessage))
+            {
+                var error = CreatePathNotFoundError(objectToApplyTo, operation.path, operation, errorMessage);
```
rynowak: Is this the right semantic for a `test`? it seems odd that you can't test against a non-existent value.

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -365,6 +366,37 @@ public void Copy(Operation operation, object objectToApplyTo)
...
+                ErrorReporter(error);
+                return;
+            }
+
+            if (!adapter.TryTest(target, parsedPath.LastSegment, ContractResolver, operation.value, out errorMessage))
+            {
+                var error = CreateOperationFailedError(objectToApplyTo, operation.path, operation, errorMessage);
```
rynowak: ???

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -365,6 +366,37 @@ public void Copy(Operation operation, object objectToApplyTo)
             }
         }
 
+        public void Test(Operation operation, object objectToApplyTo)
```
rynowak: Consider copying the relevant portions of the spec here like the others. At least they will all be wacky the same way.

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
```diff
@@ -126,6 +127,47 @@ public class DictionaryAdapter<TKey, TValue> : IAdapter
...
+            var contract = (JsonDictionaryContract)contractResolver.ResolveContract(target.GetType());
+            var key = contract.DictionaryKeyResolver(segment);
+            var dictionary = (IDictionary<TKey, TValue>)target;
+
+            if (!TryConvertKey(key, out var convertedKey, out errorMessage))
+            {
+                return false;
+            }
+
+            // As per JsonPatch spec, the target location must exist for remove to be successful
```
rynowak: ???

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
```diff
@@ -126,6 +127,47 @@ public class DictionaryAdapter<TKey, TValue> : IAdapter
...
+
+            if (!TryConvertValue(value, out var convertedValue, out errorMessage))
+            {
+                return false;
+            }
+
+            var currentValue = dictionary[convertedKey];
+            if (!(JsonConvert.SerializeObject(currentValue) == JsonConvert.SerializeObject(convertedValue)))
+            {
+                errorMessage = Resources.FormatValueNotEqualToTestValue(currentValue, value, segment);
```
rynowak: Is this an error?

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
```diff
@@ -126,6 +127,47 @@ public class DictionaryAdapter<TKey, TValue> : IAdapter
...
+                return false;
+            }
+
+            if (!TryConvertValue(value, out var convertedValue, out errorMessage))
+            {
+                return false;
+            }
+
+            var currentValue = dictionary[convertedKey];
+            if (!(JsonConvert.SerializeObject(currentValue) == JsonConvert.SerializeObject(convertedValue)))
```
rynowak: We might want to use `JObject` here. Also, the contract resolver needs to be involved somehow

src/Microsoft.AspNetCore.JsonPatch/breakingchanges.netcore.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+      "TypeId": "public interface Microsoft.AspNetCore.JsonPatch.Adapters.IObjectAdapter",
+      "MemberId": "System.Void Test(Microsoft.AspNetCore.JsonPatch.Operations.Operation operation, System.Object objectToApplyTo)",
+      "Kind": "Addition"
+    }
+  ]
```
rynowak: When you see something like this pop up - we need to discuss and get approval - or find an alternative.

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
jbagga: If it cannot find that segment, how can it test against its value? Don't think the spec defines the exact behavior in that case.

https://tools.ietf.org/html/rfc6902#page-7

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
jbagga: What about this is the problem? I thought it just propagates the error from TryTest - "path not found" or "test operation failed"

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
jbagga: Yes

src/Microsoft.AspNetCore.JsonPatch/breakingchanges.netcore.json
jbagga: @Eilon knew about this. But I should look into the non-breaking change way of adding this you suggested above. Thanks

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
Eilon: Yeah the spec isn't super clear about it. This is the closest thing I can find: https://tools.ietf.org/html/rfc6902#appendix-A.12

> A.12.  Adding to a Nonexistent Target
>
>   An example target JSON document:
>
>   { "foo": "bar" }
>
>   A JSON Patch document:
>
>   [
     { "op": "add", "path": "/baz/bat", "value": "qux" }
   ]
>
>   This JSON Patch document, applied to the target JSON document above,
   would result in an error (therefore, it would not be applied),
   because the "add" operation's target location that references neither
   the root of the document, nor a member of an existing object, nor a
   member of an existing array.

Which would lead me to believe that much like doing an `add` to something that doesn't exist is a failure (i.e. abort the whole operation), then doing a `test` in a similar case would also be a failure. (Here's where failures are discussed: https://tools.ietf.org/html/rfc6902#section-5.)

src/Microsoft.AspNetCore.JsonPatch/Adapters/IObjectAdapter.cs
Eilon: Yeah a 2nd interface might be the best choice here. Too bad we got it wrong in v1 😦

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
Eilon: Per my other comment about what's allowed, I think this is correct, no?

src/Microsoft.AspNetCore.JsonPatch/Internal/ObjectVisitor.cs
```diff
@@ -42,14 +39,14 @@ public bool TryVisit(ref object target, out IAdapter adapter, out string errorMe
...
-                adapter = SelectAdapater(target);
+                adapter = SelectAdapter(target);
             }
 
             errorMessage = null;
             return true;
         }
 
-        private IAdapter SelectAdapater(object targetObject)
+        private IAdapter SelectAdapter(object targetObject)
```
Eilon: ⬇️  https://en.wikipedia.org/wiki/Lord_Byron

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
jbagga: I need to change it to say "test" instead of "remove". My bad

Eilon: Ah didn't realize this was a about a copy pasta typo... I think the comment is otherwise correct.

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
rynowak: Yeah, the section about test talks about comparing values. I think they mean to say that there must be a value present in the target - but I would like it better if it was spelled out.

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
rynowak: I think the high level thing we need to deal with here is - **is it a model state error when a test fails?**

jbagga: The json patch request should be abandoned if the test fails. It should no-op and return some sort of an error.

```
   An example target JSON document:
   {
     "/": 9,
     "~1": 10
   }

   A JSON Patch document:

   [
     {"op": "test", "path": "/~01", "value": "10"}
   ]
```
> This results in an error, because the test fails.

src/Microsoft.AspNetCore.JsonPatch/Internal/DictionaryAdapterOfTU.cs
jbagga: The key being used has already been resolved with `contract.DictionaryKeyResolver(segment)` https://github.com/aspnet/JsonPatch/pull/114/files/8763258935a6eb2b7c151cda519884bf6f8dd041#diff-eba89d20437ba32b8dca5690b028526bR138

jbagga:
@Eilon @rynowak Updated. Need to add more tests to ensure the equality comparer works as expected

src/Microsoft.AspNetCore.JsonPatch/JsonPatchDocument.cs
```diff
@@ -213,6 +231,30 @@ public void ApplyTo(object objectToApplyTo, IObjectAdapter adapter)
             }
         }
 
+        /// <summary>
+        /// Apply this JsonPatchDocument
+        /// </summary>
+        /// <param name="objectToApplyTo">Object to apply the JsonPatchDocument to</param>
+        /// <param name="adapter">IObjectAdapterWithTest instance to use when applying</param>
+        public void ApplyTo(object objectToApplyTo, IObjectAdapterWithTest adapter)
```
Eilon: I think that the old `ApplyTo(object objectToApplyTo, IObjectAdapter adapter)` overload should test for the `adapter` param implementing `IObjectAdapterWithTest`, and if so, call the new overload so that legacy compiled code can benefit from the new Test operation without having to be recompiled.

@rynowak - any additional thoughts on this?

rynowak: That seems right to me

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -392,7 +441,7 @@ public void Copy(Operation operation, object objectToApplyTo)
             var visitor = new ObjectVisitor(parsedPath, ContractResolver);
 
             var target = objectToGetValueFrom;
-            if (!visitor.TryVisit(ref target, out var adapter, out var errorMessage))
+            if (!visitor.TryVisit(ref target, adapterWithTest: out var adapter, errorMessage: out var errorMessage))
```
rynowak: I'm not thrilled with the parameter name `adapterWithTest` - it's just verbose without communicating any more. You can also notice that it doesn't match the variable name `adapter`.

src/Microsoft.AspNetCore.JsonPatch/Internal/IAdapterWithTest.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.AspNetCore.JsonPatch.Internal
+{
+    public interface IAdapterWithTest : IAdapter
```
rynowak: Why is this a separate interface? Isn't `IAdapter` internal so we're free to change it? Or did we try to be super careful?

src/Microsoft.AspNetCore.JsonPatch/Internal/ObjectVisitor.cs
```diff
@@ -42,14 +39,42 @@ public bool TryVisit(ref object target, out IAdapter adapter, out string errorMe
...
-                adapter = SelectAdapater(target);
+                adapter = SelectAdapter(target);
             }
 
             errorMessage = null;
             return true;
         }
 
-        private IAdapter SelectAdapater(object targetObject)
+        public bool TryVisit(ref object target, out IAdapterWithTest adapterWithTest, out string errorMessage)
```
rynowak: Should we just change the original `TryVisit`? this is a `.Internal` class

src/Microsoft.AspNetCore.JsonPatch/Internal/IAdapterWithTest.cs
Eilon: Oh good point about that interface. Maybe we can get away with just having one interface on the internal side?

src/Microsoft.AspNetCore.JsonPatch/Internal/ObjectVisitor.cs
jbagga: Yeah it isn't required now that we just need one `IAdapter` (internal so test can be added as an operation to it).

src/Microsoft.AspNetCore.JsonPatch/JsonPatchDocument.cs
jbagga: I don't think we need an overload. The new interface implements the old one so it can use the older method. The two are handled differently in `Apply` method in `Operation` (Part of my update).

jbagga:
@Eilon @rynowak Updated

src/Microsoft.AspNetCore.JsonPatch/Operations/Operation.cs
```diff
@@ -58,7 +58,15 @@ public void Apply(object objectToApplyTo, IObjectAdapter adapter)
...
-                    throw new NotSupportedException(Resources.TestOperationNotSupported);
+                    if (adapter is IObjectAdapterWithTest adapterWithTest)
+                    {
+                        adapterWithTest.Test(this, objectToApplyTo);
+                        break;
+                    }
+                    else
+                    {
+                        throw new NotSupportedException(Resources.TestOperationNotSupported);
+                    }
```
rynowak: 👍

src/Microsoft.AspNetCore.JsonPatch/Internal/PocoAdapter.cs
```diff
@@ -160,7 +161,7 @@ public class PocoAdapter : IAdapter
             }
 
             var currentValue = jsonProperty.ValueProvider.GetValue(target);
-            if (!Equals(currentValue, convertedValue))
```
rynowak: Let's make sure we have good tests for some interesting cases:
- string
- number
- object
- array

jbagga: The unit tests for all the adapters cover these. I am adding integration tests as part of the clean up for #105

PR #116, opened 2017-10-19
Throw JsonSerializationException from converters

jbagga:
See https://github.com/aspnet/Entropy/pull/268

The converters need to send a JSON.NET exception and not a JsonPatchException so the MVC formatters can correctly propagate it.

cc @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.JsonPatch/Converters/JsonPatchDocumentConverter.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/Converters/JsonPatchDocumentConverter.cs
@@ -57,7 +57,7 @@ namespace Microsoft.AspNetCore.JsonPatch.Converters
             }
             catch (Exception ex)
             {
-                throw new JsonPatchException(Resources.FormatInvalidJsonPatchDocument(objectType.Name), ex);
+                throw new JsonSerializationException(Resources.FormatInvalidJsonPatchDocument(), ex);
             }
         }
 
--- a/src/Microsoft.AspNetCore.JsonPatch/Converters/JsonPatchDocumentConverter.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/Converters/JsonPatchDocumentConverter.cs
@@ -57,7 +57,7 @@ namespace Microsoft.AspNetCore.JsonPatch.Converters
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.JsonPatch/Converters/JsonPatchDocumentConverter.cs
```diff
@@ -57,7 +57,7 @@ public override bool CanConvert(Type objectType)
             }
             catch (Exception ex)
             {
-                throw new JsonPatchException(Resources.FormatInvalidJsonPatchDocument(objectType.Name), ex);
+                throw new JsonSerializationException(Resources.FormatInvalidJsonPatchDocument(), ex);
```
rynowak: nit: you can use `Resources.InvalidJsonPatchDocument` since there are no parameters.

src/Microsoft.AspNetCore.JsonPatch/Resources.resx
```diff
@@ -142,7 +142,7 @@
     <value>The path segment '{0}' is invalid for an array index.</value>
   </data>
   <data name="InvalidJsonPatchDocument" xml:space="preserve">
-    <value>The type '{0}' was malformed and could not be parsed.</value>
+    <value>The JsonPatchDocument was malformed and could not be parsed.</value>
```
rynowak: `The JSON patch document was malformed and could not be parsed.`

The error message refers to the input, not to the specific type.

test/Microsoft.AspNetCore.JsonPatch.Test/ObjectAdapterTests.cs
```diff
@@ -756,19 +756,35 @@ public void DeserializationMustWorkWithoutEnvelope()
...
                 var deserialized
                     = JsonConvert.DeserializeObject<JsonPatchDocument<SimpleObject>>(serialized);
             });
 
-            Assert.Equal("The type 'JsonPatchDocument`1' was malformed and could not be parsed.", exception.Message);
+            Assert.Equal("The JsonPatchDocument was malformed and could not be parsed.", exception.Message);
+        }
+
+        [Fact]
+        public void DeserializationMustFailWithEnvelope()
```
rynowak: These test names are really silly. Can you come up with something more in line with our guidelines.

jbagga:
@rynowak Updated

PR #117, opened 2017-10-19
Reorganize tests

jbagga:
Part I for #105 

This is the first PR in process of removing redundancy and adding structure to the tests


jbagga:
cc @Eilon

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/AnonymousObjectIntegrationTest.cs
```diff
@@ -0,0 +1,242 @@
...
+
+            var serialized = JsonConvert.SerializeObject(patchDocument);
+            var deserialized = JsonConvert.DeserializeObject<JsonPatchDocument>(serialized);
+
+            var exception = Assert.Throws<JsonPatchException>(() =>
+            {
+                deserialized.ApplyTo(targetObject);
+            });
+            Assert.Equal(
+                string.Format("The index value provided by path segment '{0}' is out of bounds of the array size.", "4"),
```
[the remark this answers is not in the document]
jbagga: I just moved some of the old tests to this class. I can fix the many instances I've seen of this in the next PR. I'll add this to the issue list. Thanks!

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/AnonymousObjectIntegrationTest.cs
```diff
@@ -0,0 +1,242 @@
...
+using System;
+using System.Collections.Generic;
+using System.Dynamic;
+using Microsoft.AspNetCore.JsonPatch.Exceptions;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Microsoft.AspNetCore.JsonPatch.Test.IntegrationTests
+{
+    public class AnonymousObjectIntegrationTest
```
rynowak: Yay

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/AnonymousObjectIntegrationTest.cs
```diff
@@ -0,0 +1,242 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Dynamic;
+using Microsoft.AspNetCore.JsonPatch.Exceptions;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Microsoft.AspNetCore.JsonPatch.Test.IntegrationTests
```
rynowak: In general don't put `.Test` in namespaces

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/AnonymousObjectIntegrationTest.cs
```diff
@@ -0,0 +1,242 @@
...
+{
+    public class AnonymousObjectIntegrationTest
+    {
+        [Fact]
+        public void ShouldAddToList_WithDifferentCase()
```
rynowak: Wierd name

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/AnonymousObjectIntegrationTest.cs
```diff
@@ -0,0 +1,242 @@
...
+        {
+            var targetObject = new
+            {
+                IntegerList = new List<int>() { 1, 2, 3 }
+            };
+
+            // create patch
```
rynowak: Comments like this aren't really needed they are just obvious

rynowak: I also realize that this is just the copy-paste part so I'm going to stop commenting

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/AnonymousObjectIntegrationTest.cs
```diff
@@ -0,0 +1,242 @@
...
+            var patchDocument = new JsonPatchDocument();
+            patchDocument.Add("IntegerList/4", 4);
+
+            var serialized = JsonConvert.SerializeObject(patchDocument);
+            var deserialized = JsonConvert.DeserializeObject<JsonPatchDocument>(serialized);
```
rynowak: We should blanket-remove all of these silliness. Add it to the list

jbagga: The fourth item is this type of clean up https://github.com/aspnet/JsonPatch/issues/105

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/AnonymousObjectIntegrationTest.cs
```diff
@@ -0,0 +1,242 @@
...
+            var patchDocument = new JsonPatchDocument();
+            patchDocument.Add("IntegerList/-", 4);
+
+            var serialized = JsonConvert.SerializeObject(patchDocument);
+            var deserialized = JsonConvert.DeserializeObject<JsonPatchDocument>(serialized);
+
+            deserialized.ApplyTo(targetObject);
+
+            Assert.Equal(new List<int>() { 1, 2, 3, 4 }, targetObject.IntegerList);
+        }
```
rynowak: I'm not sure there's much value in having tests that use a strategy like this. 

For instance these are the 'anonymous object' tests, but 5 that are related to the behavior of lists. If anything those should be the 'list tests'. If they are duplicated we can remove them.

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/AnonymousObjectIntegrationTest.cs
```diff
@@ -0,0 +1,242 @@
...
+
+            var newGuid = Guid.NewGuid();
+            // create patch
+            var patchDocument = new JsonPatchDocument();
+            patchDocument.Replace("nestedobject/GuidValue", newGuid);
+
+            patchDocument.ApplyTo(targetObject);
+
+            Assert.Equal(newGuid, targetObject.NestedObject.GuidValue);
+        }
```
rynowak: I think that most of these tests don't add unique value if the **pivot** in this case is anonymously-typed objects. anon-t objects behave like POCO's with `get; private set;` properties. 

It seems like we need 4 tests here.
1. We can get a property's value
2. We cannot set a property's value
3. We cannot add a property
4. We cannot remove a property

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/DictionaryIntegrationTest.cs
```diff
@@ -0,0 +1,225 @@
...
+            var address = actualValue1.Address;
+            Assert.NotNull(address);
+            Assert.Equal("Bellevue", address.City);
+        }
+
+        class Class9
+        {
+            public List<string> StringList { get; set; } = new List<string>();
+        }
+    }
```
rynowak: These need some tests related to non-string keys

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/DictionaryIntegrationTest.cs
```diff
@@ -0,0 +1,225 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Microsoft.AspNetCore.JsonPatch
```
rynowak: Should `.IntegrationTests` be in the namespace?

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/DictionaryIntegrationTest.cs
rynowak: Most of these are good though

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/DynamicObjectIntegrationTest.cs
```diff
@@ -287,13 +287,13 @@ public void TestPropertyValue_FromListToNonList_InNestedTypedObject_InDynamicObj
...
-            var patchDoc = new JsonPatchDocument();
-            patchDoc.Test("Nested/IntegerList/0", 2);
+            var patchDocument = new JsonPatchDocument();
+            patchDocument.Test("Nested/IntegerList/0", 2);
 
             // Act
             var exception = Assert.Throws<JsonPatchException>(() =>
             {
-                patchDoc.ApplyTo(dynamicTestObject);
+                patchDocument.ApplyTo(dynamicTestObject);
```
rynowak: The dynamic tests seem to have gone wayyy in depth including things like replace and move where a lot of the others don't. I'm not sure if that's good or bad.

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/ExpandoObjectIntegrationTest.cs
```diff
@@ -0,0 +1,841 @@
...
+            var deserialized = JsonConvert.DeserializeObject<JsonPatchDocument>(serialized);
+
+            deserialized.ApplyTo(obj);
+
+            Assert.Equal(1, obj.NewInt);
+            Assert.Equal(1, obj.Test);
+        }
+
+        [Fact]
+        public void AddNewProperty_ToExpandoOject_InTypedObject()
```
rynowak: I think that there are way too many of these kinds of tests. Putting the target object at another level of testing doesn't make it a valuable test. A few of these make sense.

rynowak:
I think that what you've done is create a pretty reasonable organization based on what's there. 

There are still a lot of non-meaningful tests - these are the ones like "replace an item in a list in a dynamic object in a dictionary" and the 345 other variations. It might be good to go through each class separately and just see what tests in that class are:
- Related to the core concept being tested (dictionaries, dynamic, etc)
- Not covered by unit tests
- Have a reason to exist

To me, most of the features in JsonPatch are not related. Lists and dynamic are not connected. So the right scale is O(n) tests for list and O(m) tests for dynamic other than O(m*n) tests for both combined.

jbagga:
@rynowak @Eilon Updated

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPatchDocumentJsonPropertyAttributeTest.cs
```diff
@@ -15,57 +15,66 @@ public class JsonPatchDocumentJsonPropertyAttributeTest
         [Fact]
         public void Add_ToRoot_OfListOfObjects_AtEndOfList()
         {
-            var patchDoc = new JsonPatchDocument<List<JsonPropertyObject>>();
-            patchDoc.Add(p => p, new JsonPropertyObject());
+            // Arrange
+            var patchDocument = new JsonPatchDocument<List<JsonPropertyObject>>();
+            patchDocument.Add(p => p, new JsonPropertyObject());
+
+            var serialized = JsonConvert.SerializeObject(patchDocument);
```
rynowak: >var serialized = JsonConvert.SerializeObject(patchDocument); [](start = 12, length = 60)

There are more cases for #5 in this file. You can probably do a find for `SerializeObject(patchDocument)` and find them all

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPatchDocumentJsonPropertyAttributeTest.cs
```diff
@@ -80,142 +89,157 @@ public void Add_WithExpressionOnStringProperty_FallsbackToPropertyName_WhenJsonP
...
 
-            deserialized.ApplyTo(doc);
-
-            Assert.Null(doc.Name);
+            // Assert
+            Assert.Null(targetObject.Name);
         }
 
         [Fact]
         public void Add_OnApplyFromJson_RespectsInheritedJsonPropertyNameOnJsonDocument()
```
rynowak: >Add_OnApplyFromJson_RespectsInheritedJsonPropertyNameOnJsonDocument [](start = 20, length = 67)

I don't think we need as many of these tests as there are. This is just re-testing how JsonContract behaves. We probably need 1-2 cases just to show that the end to end works

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/AnonymousObjectIntegrationTest.cs
```diff
@@ -0,0 +1,215 @@
...
+                patchDocument.ApplyTo(targetObject);
+            });
+
+            // Assert
+            Assert.Equal("The property at path 'Test' could not be updated.",
+                exception.Message);
+        }
+
+        [Fact]
+        public void ReplaceGuid_InNestedExpandoObject()
```
rynowak: >ReplaceGuid_InNestedExpandoObject [](start = 20, length = 33)

This has nothing to do with this class

test/Microsoft.AspNetCore.JsonPatch.Test/IntegrationTests/NestedObjectIntegrationTest.cs
```diff
@@ -0,0 +1,349 @@
...
+            var patchDocument = new JsonPatchDocument<SimpleObjectWithNestedObject>();
+            patchDocument.Replace(o => o.SimpleObject.StringProperty, "B");
+            patchDocument.Replace(o => o.SimpleObject.DecimalValue, 12);
+            patchDocument.Replace(o => o.SimpleObject.DoubleValue, 12);
+            patchDocument.Replace(o => o.SimpleObject.FloatValue, 12);
+            patchDocument.Replace(o => o.SimpleObject.IntegerValue, 12);
+
+            // serialize & deserialize
+            var serialized = JsonConvert.SerializeObject(patchDocument);
+            var deserizalized = JsonConvert.DeserializeObject<JsonPatchDocument<SimpleObjectWithNestedObject>>(serialized);
```
rynowak: We don't need this. Looks like there are a few 'with serialization' tests

jbagga:
@rynowak Updated

PR #125, opened 2017-11-06
allow paths that contain '.'

### LINKED ISSUE #124: Paths containing '.' trigger JsonPatchException
In [/src/Microsoft.AspNetCore.JsonPatch/Internal/PathHelpers.cs](https://github.com/aspnet/JsonPatch/blob/dev/src/Microsoft.AspNetCore.JsonPatch/Internal/PathHelpers.cs):

`if (path.Contains(".") || path.Contains("//") || path.Contains(" ") || path.Contains("\\"))
{
    throw new JsonPatchException(Resources.FormatInvalidValueForPath(path), null); 
}`

The spec at jsonpatch.com doesn't ever specify that '.' characters are invalid or must be escaped.

arerlend:
fixes #124 

JSON Patch spec as found at jsonpatch.com never specifies that '.' character are invalid.  Several major REST APIs with HTTP Patch endpoints use paths that contain '.' characters.

### DIFF
--- a/src/Microsoft.AspNetCore.JsonPatch/Internal/PathHelpers.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/Internal/PathHelpers.cs
@@ -13,7 +13,7 @@ namespace Microsoft.AspNetCore.JsonPatch.Internal
             // absolutely necessary, but it allows us to already catch mistakes
             // on creation of the patch document rather than on execute.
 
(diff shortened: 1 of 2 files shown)


arerlend:
@Eilon

rynowak:
Looks good to me also. Merging.

I think the intent from the original code was to catch errors from accidentally creating a JSON patch path from a C# expression like `x => x.Foo.Bar`. It should create `/Foo/Bar` not `/Foo.Bar`.

rynowak:
@arerlend Thanks!

PR #130, opened 2018-02-27
Review to support overrides within JsonPatch

jmudavid:
NOT READY FOR MERGE - ONLY REVIEW

@Eilon,
Here are the changes I am proposing.  I don't believe there are any breaking changes to the public interfaces.  
- Re-scoped many private methods to protected
- Added virtual to many public methods to allow for overrides
- Added Interfaces for IObjectVisitor and ParsedPath
- Implemented the new interfaces in the existing concrete classes

Here is why I am doing so if you are interested in the use case.  We are implementing the SCIM protocol which has its own protocol for patch (https://tools.ietf.org/html/rfc7644#page-32) based on the standard Json Patch functionality.  The main differences I needed these changes to support was the ability to select from an array of items not be index, but by SCIM filters.  With these changes I am able to add Json Patch adapters that follow the SCIM filters and support the ParsedPath and ObjectVisitor functionality needed for the rest of JsonPatch to work as it does.  

Please let me know your thoughts on next steps.  I know my branch is currently a few commits behind master but that should be any easy fix.

Thanks,
Dave


Eilon:
@jmudavid thanks for sending this.

@mkArtakMSFT - can you get the right group of people to evaluate this PR for issue https://github.com/aspnet/Home/issues/2816?

jmudavid:
Closing and opening a new PR using a `IAdapterFactory`

PR #131, opened 2018-02-27
Allow whitespace and backslash in path

ajaybhargavb:
Fixes https://github.com/aspnet/Home/issues/2876

From looking at the spec https://tools.ietf.org/html/rfc6902#section-4 which points to https://tools.ietf.org/html/rfc6901, whitespace and backslash in path should be allowed.

FYI @mkArtakMSFT

### DIFF
--- a/src/Microsoft.AspNetCore.JsonPatch/Internal/PathHelpers.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/Internal/PathHelpers.cs
@@ -2,23 +2,24 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.JsonPatch.Exceptions;
+using System;
 
 namespace Microsoft.AspNetCore.JsonPatch.Internal
 {
     internal static class PathHelpers
     {
-        internal static string NormalizePath(string path)
+        internal static string ValidateAndNormalizePath(string path)
         {
             // check for most common path errors on create.  This is not
             // absolutely necessary, but it allows us to already catch mistakes
             // on creation of the patch document rather than on execute.
 
-            if (path.Contains("//") || path.Contains(" ") || path.Contains("\\"))
+            if (path.Contains("//"))
             {
                 throw new JsonPatchException(Resources.FormatInvalidValueForPath(path), null); 
             }
 
-            if (!(path.StartsWith("/")))
+            if (!path.StartsWith("/", StringComparison.Ordinal))
             {
                 return "/" + path;
             }
--- a/test/Microsoft.AspNetCore.JsonPatch.Test/JsonPatchDocumentJsonPropertyAttributeTest.cs
+++ b/test/Microsoft.AspNetCore.JsonPatch.Test/JsonPatchDocumentJsonPropertyAttributeTest.cs
@@ -25,6 +25,23 @@ namespace Microsoft.AspNetCore.JsonPatch
         }
 
         [Fact]
+        public void Add_RespectsJsonPropertyAttribute_WithDotWhitespaceAndBackslashInName()
+        {
+            // Arrange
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.JsonPatch/Internal/PathHelpers.cs
```diff
@@ -13,7 +13,7 @@ internal static string NormalizePath(string path)
             // absolutely necessary, but it allows us to already catch mistakes
             // on creation of the patch document rather than on execute.
 
-            if (path.Contains("//") || path.Contains(" ") || path.Contains("\\"))
+            if (path.Contains("//"))
```
Eilon: Since you're here, can you change both the `.Contains()` and the `.StartsWith` below to either:
1. Use the `char`-based check
2. If not possible, pass in the explicit `StringComparison.Ordinal`

test/Microsoft.AspNetCore.JsonPatch.Test/JsonPatchDocumentJsonPropertyAttributeTest.cs
```diff
@@ -24,6 +24,23 @@ public void Add_RespectsJsonPropertyAttribute()
             Assert.Equal("/AnotherName", pathToCheck);
         }
 
+        [Fact]
+        public void Add_RespectsJsonPropertyAttribute_WithDotWhitespaceAndBackslashInName()
+        {
+            // Arrange
+            var obj = new JsonPropertyObjectWithStrangeNames();
```
pranavkm: Lol

src/Microsoft.AspNetCore.JsonPatch/Internal/PathHelpers.cs
pranavkm: Might be worthwhile to change the name to `ValidateAndNormalizePath`. Also the `StartsWith` below needs a `StringComparison.Ordinal`

ajaybhargavb:
> Maybe include the information that it cannot contain "//".

@jbagga can't do that. The same error message is also used in other places like https://github.com/aspnet/JsonPatch/blob/dev/src/Microsoft.AspNetCore.JsonPatch/Internal/ParsedPath.cs#L62

src/Microsoft.AspNetCore.JsonPatch/Internal/PathHelpers.cs
ajaybhargavb: Updated `StartsWith`. Going to leave `Contains` as is because I think it is cleaner than doing `IndexOf` with StringComparison.Ordinal.

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.JsonPatch/Internal/PathHelpers.cs
```diff
@@ -2,23 +2,24 @@
...
             {
-                throw new JsonPatchException(Resources.FormatInvalidValueForPath(path), null); 
+                throw new JsonPatchException(Resources.FormatInvalidValueForPath(path, "//"), null); 
             }
 
-            if (!(path.StartsWith("/")))
+            if (!(path.StartsWith("/", StringComparison.Ordinal)))
```
pranavkm: `if (!path.StartsWith("/", StringComparison.Ordinal))`

src/Microsoft.AspNetCore.JsonPatch/Internal/PathHelpers.cs
Eilon: Ah, didn't realize that `.Contains()` always does ordinal case-sensitive, so this one is ok.

pranavkm: `Contains` does the right thing - https://referencesource.microsoft.com/#mscorlib/system/string.cs,2129

PR #135, opened 2018-05-11
Allow adapters to be inherited and extended

jmudavid:
- Added AdapterFactory and IAdapterFactory interface
- Updated all built in Adapters to contain virtual methods and protected members so they can be extended
- Updated ObjectAdapter to accept an AdapterFactory as part of the constructor.  If not provided, the default AdapterFactory is used
- Updated the ApplyTo methods in the JsonPatchDocument to use the default AdapterFactory when an ObjectAdapter is not provided
- Added an ApplyTo method which allows a custom AdapterFactory and supports the logErrorAction method

### DIFF
--- a/src/Microsoft.AspNetCore.JsonPatch/Adapters/AdapterFactory.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/Adapters/AdapterFactory.cs
@@ -1,0 +1,39 @@
+﻿using Microsoft.AspNetCore.JsonPatch.Internal;
+using Newtonsoft.Json.Serialization;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.JsonPatch.Adapters
+{
+    /// <summary>
+    /// The default AdapterFactory to be used for resolving <see cref="IAdapter"/>.
+    /// </summary>
+    public class AdapterFactory : IAdapterFactory
+    {
+        /// <inheritdoc />
+        public virtual IAdapter Create(object target, IContractResolver contractResolver)
+        {
+            var jsonContract = contractResolver.ResolveContract(target.GetType());
+
+            if (target is IList)
+            {
+                return new ListAdapter();
+            }
+            else if (jsonContract is JsonDictionaryContract jsonDictionaryContract)
+            {
+                var type = typeof(DictionaryAdapter<,>).MakeGenericType(jsonDictionaryContract.DictionaryKeyType, jsonDictionaryContract.DictionaryValueType);
+                return (IAdapter)Activator.CreateInstance(type);
+            }
+            else if (jsonContract is JsonDynamicContract)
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.JsonPatch/Adapters/AdapterFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    /// <summary>
+    /// The default AdapterFactory to be used for resolving <see cref="IAdapter"/>.
+    /// </summary>
+    public class AdapterFactory : IAdapterFactory
+    {
+        /// <inheritdoc />
+        public virtual IAdapter Create(object target, IContractResolver contractResolver)
+        {
+            var jsonContract = contractResolver.ResolveContract(target.GetType());
```
rynowak: Can you please add argument validation here? (throw `ArgumentNullException`) for both of these if null.

src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -18,17 +18,37 @@ public class ObjectAdapter : IObjectAdapterWithTest
...
+
         /// <summary>
         /// Gets or sets the <see cref="IContractResolver"/>.
         /// </summary>
         public IContractResolver ContractResolver { get; }
 
+        /// <summary>
+        /// Gets or sets the <see cref="IAdapterFactory"/>
+        /// </summary>
+        public IAdapterFactory AdapterFactory { get; set; }
```
rynowak: Does this need to be publicly settable since it's passed in the constructor?

jmudavid: @HappyNomad and I had discussed this earlier.  I tend to leave them settable for flexibility and habits kicked in. Removed.

jmudavid:
@rynowak Changes made... fingers crossed...  Thanks again for your help on all of this.

As a side question, assuming this gets merged, are there preview NuGets available that we can use until 2.2 is released, or do we need to continue using our forks?  Thanks!

rynowak:
Looks awesome!

Our CI builds are available on https://dotnet.myget.org/gallery/dotnet-core

jmudavid:
@rynowak.
It looks like https://dotnet.myget.org/gallery/dotnet-core is still focusing on 2.1 and not 2.2, is that right?  The latest version is 2.1.0-rtm-30765 and doesn't appear to have these changes.  Do I just need to wait until 2.1 is finalized, or is there another place I need to look for the build.

Thanks!

rynowak:
I'll check on this.

rynowak:
We have definitely had builds pass since then... 🤔 

I think I gave you the wrong feed. Try this one... https://dotnet.myget.org/gallery/aspnetcore-dev

PR #145, opened 2018-09-26
(Re-submission) Improve conformance of replace operations to spec

TAGC:
Re-submission of #111 by request.

---

This ensures that JSON patch "replace" operations are functionally
equivalent to "remove" operations followed by "add" operations at the
same path, as RFC 6902 specifies.

Addresses #110

### DIFF
--- a/src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
@@ -96,6 +96,11 @@ namespace Microsoft.AspNetCore.JsonPatch.Internal
                 return false;
             }
 
+            if (!TryRemove(target, segment, contractResolver, out errorMessage))
+            {
+                return false;
+            }
+
             if (!TrySetDynamicObjectProperty(target, contractResolver, segment, convertedValue, out errorMessage))
             {
                 return false;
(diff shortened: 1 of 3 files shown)


mkArtakMSFT:
Thanks for your effort, @TAGC.
@rynowak, @pranavkm can you please review this?Thanks!

src/Microsoft.AspNetCore.JsonPatch/Internal/DynamicObjectAdapter.cs
```diff
@@ -96,6 +96,11 @@ public class DynamicObjectAdapter : IAdapter
                 return false;
             }
 
+            if (!TryRemove(target, segment, contractResolver, out errorMessage))
```
rynowak: Are there any real cases where `TryRemove` will fail? What are the semantics of the spec when you have a replace operation but there is "nothing to remove"?

TAGC: According to the spec, `replace` should function as a `remove` followed by an `add`. The spec says for the `remove` op:

> The target location MUST exist for the operation to be successful.

So I believe that `replace` will fail if the target location does not already exist. `TryRemove` looks like it verifies the target location exists and will fail if it doesn't.

TAGC: I think I misunderstood what you meant actually. If you mean what should happen if a property exists within the object graph at the location specified by the patch operation, but it's set to null or something like that - I believe the replace operation should still succeed, in that I can't find anything in the spec that explicitly rules against that. I think the present implementation allows for that as well - `TryConvertValue` looks like it succeeds for properties that are reference types and are set to null.

rynowak: ok

PR #151, opened 2018-11-08
Fix aspnet/AspNetCore#3559

Iamcerba:
Fixes https://github.com/aspnet/AspNetCore/issues/3559

Json Patch: System.NullReferenceException while trying to use copy operation from property with null value.

Note: Even if we are removing "IsNullableType" check which is not necessary in this place, we will proceed with this check later in ConversionResultProvider.ConvertTo#12 method.

### DIFF
--- a/src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
+++ b/src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
@@ -229,7 +229,7 @@ namespace Microsoft.AspNetCore.JsonPatch.Adapters
             if (TryGetValue(operation.from, objectToApplyTo, operation, out var propertyValue))
             {
                 // Create deep copy
-                var copyResult = ConversionResultProvider.CopyTo(propertyValue, propertyValue.GetType());
+                var copyResult = ConversionResultProvider.CopyTo(propertyValue, propertyValue?.GetType());
                 if (copyResult.CanBeConverted)
                 {
                     Add(operation.path,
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -229,7 +229,7 @@ public void Copy(Operation operation, object objectToApplyTo)
             if (TryGetValue(operation.from, objectToApplyTo, operation, out var propertyValue))
             {
                 // Create deep copy
-                var copyResult = ConversionResultProvider.CopyTo(propertyValue, propertyValue.GetType());
+                var copyResult = ConversionResultProvider.CopyTo(propertyValue, propertyValue?.GetType());
```
pranavkm: Wouldn't this null ref: https://github.com/aspnet/JsonPatch/blob/1a1b34194b387175ef1bc12d2af6480ae1ba9c0c/src/Microsoft.AspNetCore.JsonPatch/Internal/ConversionResultProvider.cs#L18-L22

Iamcerba: @pranavkm 
In the described scenario, no. If you check the only usage "CopyTo" method, typeToConvertTo will be null only in case of value == null and this statement will never execute. In case of "ConvertTo" method typeToConvertTo always populated from Path segment and should have value in any case.

pranavkm:
Thanks for the PR!

